PR #48, opened 2019-02-01
Association transactions CSV export

mhvis:


UserDetails/views_association.py
```diff
@@ -42,6 +44,39 @@ def get_context_data(self, **kwargs):
...
+
+    def get(self, request, *args, **kwargs):
+        response = HttpResponse(content_type='text/csv')
+        response['Content-Disposition'] = 'attachment; filename="association_transactions.csv"'
+        csv_writer = csv.writer(response)
+        # Write header
+        csv_writer.writerow(['Created on', 'Executed on', 'Source type', 'Source name', 'Source e-mail', 'Target type',
+                             'Target name', 'Target e-mail', 'Amount', 'Description'])
+        # Write transactions
+        for t in AbstractTransaction.get_all_transactions(association=self.association):
```
DutcherNL: I think 'FixedTransaction.get_all_transactions(association=self.association)' might be more suited here. Exporting a file kinda treats all transactions as immutable, but because some transactions are still editable (and therefore aren't), it could be better to leave those out of export (or implement an option whether those should be included.

mhvis: Gefixt

mhvis:
Closed because outdated (not merging into master)

PR #53, opened 2019-02-03
Remove UserDetail, remove AssociationDetail

mhvis:
This originated as a bugfix for a bug that happened when creating a new association and not providing all info. It ended up being a removal of the UserDetail and AssociationDetail models.


UserDetails/models.py
```diff
@@ -35,39 +41,8 @@ def can_access_back(self):
 
 
 class Association(Group):
-    class Meta:
```
DutcherNL: Dat moment dat je ziet dat Association nog een proxy is terwijl dat al lang niet handig meer is XD Goede fix.

PR #56, opened 2019-02-07
Migrations bugfix

mhvis:
I was not able to install it due to a bug in the migration files, this was the fix for it. Not sure yet if it has unforeseen consequences.


mhvis:
I'm going to merge it now directly. Sorry for this. I'd like it in master so that I can continue from these migrations in the other pull requests and also because it's needed for the deployment.

PR #57, opened 2019-02-07
allauth

mhvis:
I've used allauth in two other projects, this could maybe be nice to incorporate here as well? It has nice e-mail management where users can have multiple e-mails and addresses can be verified. It also includes password recovery via mail and supports OpenID providers and can thus also be used to support multiple association login links.

https://github.com/pennersr/django-allauth


mhvis:
Todo:

- [ ] Use allauth for Quadrivium OpenID Connect instead of the current library
- [ ] E-mail management
- [ ] Password change
- [ ] Direct log out

mhvis:
Would we like to require the e-mail to be verified before the user may use her account?

mhvis:
This PR now has e-mail management with e-mail verification, password change and password reset. The e-mail verification is currently optional but can be set to mandatory in `settings.py`. Only thing missing is that the Quadrivium login still uses the other library while it can also be implemented using this library. I'll put that in a separate PR.

The user settings page is changed quite a lot to incorporate the e-mail and password changes. It is also renamed to 'profile' to conform to the library and Django docs convention ([.e.g.](https://docs.djangoproject.com/en/2.1/ref/settings/#login-redirect-url)). I hope you don't mind.

DutcherNL:
> Would we like to require the e-mail to be verified before the user may use her account?

I think for now let's allow usage while account is not verified to ease the use. If we notice things going wrong we could always start implementing it as a requirement.

PR #61, opened 2019-02-11
Added check on dining_time to adhere kitchen opening time

DutcherNL:
Limited dining time in forms to the kitchen use times


Dining/forms.py
```diff
@@ -53,6 +54,15 @@ def __init__(self, user, date, *args, **kwargs):
             self.fields['association'].initial = available[0].pk
             self.fields['association'].disabled = True
 
+    def clean(self, *args, **kwargs):
+        cleaned_data = super(CreateSlotForm, self).clean(*args, **kwargs)
+        # Check if the time is within time limits
+        serve_time = cleaned_data['serve_time']
+        if serve_time < settings.KITCHEN_USE_START_TIME:
+            self.add_error('serve_time', _("Kitchen can't be used this early"))
```
mhvis: Should be `raise ValidationError()` I think.

Dining/forms.py
```diff
@@ -53,6 +54,15 @@ def __init__(self, user, date, *args, **kwargs):
...
+        if serve_time > settings.KITCHEN_USE_END_TIME:
+            self.add_error('serve_time', _("Kitchen can't be used this late"))
```
mhvis: Needs `return cleaned_data` at the end to correctly override `clean()`

Dining/forms.py
```diff
@@ -76,6 +86,15 @@ class Meta:
...
         fields = ['serve_time', 'min_diners', 'max_diners', 'sign_up_deadline', 'purchaser']
 
+    def clean(self, *args, **kwargs):
+        cleaned_data = super(DiningInfoForm, self).clean(*args, **kwargs)
+        # Check if the time is within time limits
+        serve_time = cleaned_data['serve_time']
+        if serve_time < settings.KITCHEN_USE_START_TIME:
+            self.add_error('serve_time', _("Kitchen can't be used this early"))
+        if serve_time > settings.KITCHEN_USE_END_TIME:
+            self.add_error('serve_time', _("Kitchen can't be used this late"))
```
mhvis: Needs `return cleaned_data`.

Also duplicate code, maybe better to put in model clean or MixIn or superclass.

Dining/forms.py
DutcherNL: Nope, this makes the error appear at the specefic field instead of the full form.

Dining/forms.py
DutcherNL: Used Mixin, that was a good idea.

PR #62, opened 2019-02-14
Transaction finalisation processing

DutcherNL:
Implemented finalisation of pending transactions
Reimplemented negative_since computations
Implemented dininglist signup balance checks
Implemented small corrections in dining forms


CreditManagement/migrations/0006_auto_20190213_1452.py
```diff
@@ -0,0 +1,34 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('CreditManagement', '0005_usercredit_view'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='fixedtransaction',
+            name='confirm_moment',
```
wildarch: NIT: This is usually referred to as a `timestamp` instead of a `moment`

CreditManagement/models.py
```diff
@@ -289,6 +313,26 @@ def finalise(self):
             self.delete()
             fixed_transaction.save()
 
+        return fixed_transaction
+
+    def save(self, *args, **kwargs):
+        # If no confirm moment is given, set it to the standard
+        if self.confirm_moment is None:
+            self.confirm_moment = self.order_moment + settings.MINIMUM_BALANCE
```
wildarch: What does `MINIMUM_BALANCE` do here?

CreditManagement/models.py
```diff
@@ -438,14 +492,43 @@ class UserCredit(models.Model):
...
+                                  max_digits=6)
+    balance_fixed = models.DecimalField(blank=True,
+                                        null=True,
+                                        db_column=FixedTransaction.balance_annotation_name,
+                                        decimal_places=2,
+                                        max_digits=6)
+
+    def negative_since(self):
+        """
+        Compute the date in the confirmed moments that
```
wildarch: that ...?

CreditManagement/models.py
```diff
@@ -438,14 +492,43 @@ class UserCredit(models.Model):
...
+        # Loop over all transactions from new to old, reverse its balance
+        transactions = FixedTransaction.get_all_transactions(user=self.user).order_by('-order_moment')
+        for transaction in transactions:
+            balance += transaction.amount
+            # If balance is positive now, return the current transaction date
+            if balance >= 0:
+                return transaction.order_moment
+
+        # This should not be reached, it would indicate that the starting balance was 0
+        raise Exception("Balance started as negative, negative_since could not be computed")
```
wildarch: Ideally you would make a custom Exception subclass for this

CreditManagement/models.py
```diff
@@ -438,14 +492,43 @@ class UserCredit(models.Model):
...
+
+        # Loop over all transactions from new to old, reverse its balance
+        transactions = FixedTransaction.get_all_transactions(user=self.user).order_by('-order_moment')
+        for transaction in transactions:
+            balance += transaction.amount
+            # If balance is positive now, return the current transaction date
+            if balance >= 0:
+                return transaction.order_moment
+
+        # This should not be reached, it would indicate that the starting balance was 0
```
wildarch: Don't you mean less than 0 here?

Dining/views.py
```diff
@@ -185,19 +176,48 @@ def dispatch(self, request, *args, **kwargs):
...
         available_slots = DiningList.objects.available_slots(self.date)
         if available_slots <= 0:
-            return HttpResponseForbidden('No available slots')
+            error = _("No free slots availlable")
+            messages.add_message(request, messages.ERROR, error)
+            return HttpResponseRedirect(self.reverse('day_view', kwargs={}))
+
         if len(DiningList.objects.filter(date=self.date).filter(claimed_by=self.request.user)) > 0:
-            return HttpResponseForbidden('You already have dining slot claimed today')
+            error = _("You already have dining slot claimed today")
```
wildarch: I think there is an 'a' missing here

Dining/views.py
```diff
@@ -185,19 +176,48 @@ def dispatch(self, request, *args, **kwargs):
...
+            return False
+
+        # No problems, use can add people
+        return True
+
     def get(self, request, *args, **kwargs):
         context = self.get_context_data()
 
+        # Check permissions, if user has no access to this page, reject it.
+        if not self.check_user_permission(request, request.user):
```
wildarch: I am not sure how to do it in Django, but would it be possible to invert these access control settings? Right now we accept requests by default and manually check if the user should be rejected, so the access control settings are similar to 'Accept by default, reject for certain paths'. A safer setting would be 'Reject by default, accept for certain paths'. Is this possible in Django?

Dining/views.py
```diff
@@ -239,11 +259,12 @@ def get(self, request, *args, **kwargs):
         return self.render_to_response(context)
 
     def post(self, request, *args, **kwargs):
-        # Todo: re-enable External Dining Entries
         context = self.get_context_data()
 
+        # Check permissions, if user has no access to this page, reject it.
+        if not self.check_user_permission(request, request.user):
+            return HttpResponseRedirect(self.reverse('slot_details'))
```
wildarch: No 401 or some kind of warning message?

Dining/views.py
mhvis: It could be similar to the `LoginRequiredMixin` and `DiningListMixin`.

CreditManagement/models.py
DutcherNL: That is indeed the wrong setting. It should be TRANSACTION_PENDING_DURATION  Don't know what happened here, I'll fix it.

CreditManagement/models.py
DutcherNL: "Compute the date from the balance_fixed table when the users balance has become negative"
new text

CreditManagement/models.py
DutcherNL: Now changed into RunTimeError

CreditManagement/models.py
DutcherNL: yes

Dining/views.py
DutcherNL: fixed

Dining/views.py
DutcherNL: I could write it in a mixin, but these specific checks aren't used in any other class and I see this not being used any time soon either. So placing it in a different class didn't make much sens to me tbh.
That said, I guess I could adjust it to do the checkups from the dispatch instead to apply on both 'get' and 'post'. Should I do this instead?

Dining/views.py
DutcherNL: The check_user_permissions generates warning messages. I didn't do a 401 as that doesn't help user experience imo as it basically signals a dead-end while the reason is not because the page is normally blocked, only in this specific instance (dininglist is already closed or full, i.e. an hour earlier the page was availlable)

Dining/views.py
mhvis: Using `dispatch` similar to `LoginRequiredMixin` is better so that all HTTP methods are covered. You will however need to write a separate Mixin class in this case because of the method execution ordering. Because if you add the dispatch method in this class, if you call `super().dispatch()` at the start of the method, the whole get/post riddle will be executed before the check. On the other hand, if you first check the permissions before calling the super method, the Login and Dining mixins have not yet been processed, since that also happens in the `dispatch` method and is only executed when calling the super method. The solution that comes to mind therefore is to have a separate mixin class which implements `dispatch` (does not need to extend any class) and put it after the `DiningListMixin`.

Other note: the whole `get` method can be rewritten to `get_context_data`, just like e.g. `DayView`. That is a standard Django idiom which ensures that the same template context is available for every HTTP method, get and post in this case. (Again some literature :smile: : https://stackoverflow.com/a/36950584/2373688)

DutcherNL: I understand your argument on using mixin and while it is slightly cleaner to use it in a dispatch it doesn't make sense (at least to me) to have a part of the current EntryAddView class to be placed in a different class even though that content is only applicable to this very class and this class only.

I did implement get_context_data()

PR #71, opened 2019-03-03
Bugfix createsuperuser

mhvis:
Overriding the `is_staff` property in this way seems risky and leads to side effects like the createsuperuser command not working anymore.

I think it is much cleaner to override [`AdminSite.has_permission()`](https://docs.djangoproject.com/en/2.1/_modules/django/contrib/admin/sites/#AdminSite.has_permission) but have not implemented that yet in this branch.


mhvis:
Note btw that the `is_staff` property in itself does nothing more than grant access to the admin site, it does not give permission to view or change anything in the admin site, i.e. the admin site would appear empty.

PR #72, opened 2019-03-03
Admin site changes

mhvis:
The admin site currently seems messy to me in terms of what one can and cannot do.

It should be possible for superusers to do as much as possible, except for things that break integrity but including changing user passwords and superuser/staff state. This is currently not possible in `master` but possible in this branch. Next to the superusers we could allow e.g. board members or scala treasurers to access the admin site and dynamically give them some permissions for some things they might need to view or change, but no more permissions than needed.

I propose to throw some admin site code away and recreate only the things that end up to be needed. For instance the BoardFilter and MemberOfFilter are currently useless because only superusers can use them, while this is not really meant for them.

In my view the admin site from a superuser perspective should be close to the actual models by just using mostly standard ModelAdmin instances without much modification like removing fields or making them read only. This allows for making changes which might be needed to fix things which would otherwise only be possible using direct database access. It's also really helpful during testing and debugging. From the perspective of a board member we could use the admin site for some management tasks, for instance we could use it to have them deny or accept association membership requests, but that is already possible in a dedicated view currently.


mhvis:
To elaborate, the dining add view in the admin site is currently customized with a custom field ordering and inlines but the actual audience as far as I know will only be superusers, for whom it's not needed to have these custom views. Therefore, I want to remove these customizations so that changes in a model (DiningList in this case) are immediately reflected in the admin site and do not require updates to the ModelAdmin. This makes sure that the admin site is flexible and is guaranteed to be up-to-date with the actual models.

DutcherNL:
I agree for most what you say. However, the dining list and possibly some user views should also be presented in a readable non-complete version for boards as we might give them advanced access as well. I'd like it if you won't remove those.

mhvis:
I've implemented `AdminSite.has_permission()` in this branch and PR #71 with that becomes invalid and I'll close it.

This PR can now be merged and I'll create a new PR for the other admin site changes.

PR #73, opened 2019-03-03
Added migration file creating the pending_dining_transact table

DutcherNL:
This was necessary to allow user accounts to be removed without error creation (due to table not existing)
Tried setting it as view, but it resulted in an odd error: "No such column: <>" where < > was the content of the description of all created pdt's (i.e. "DINING")


mhvis:
Your fix for `is_staff` modifies a Django private method and seems more like a patch/workaround. There might be more places in the Django code where they rely on having `is_staff` as a database field. So it really seems much cleaner to me to override `AdminSite.has_permission` as I suggested in #71.

mhvis:
Shall I implement the `has_permission` in PR #72? The second commit in this PR can then be discarded

mhvis:
I'll implement it in #71.

mhvis:
The second commit can be discarded in favor of #72, shall I do this and merge the other commit?

PR #79, opened 2019-03-04
Improved registration view

DutcherNL:
Fixed a bug in the form clean method not returning the cleaned results
Added required field display in bootstrap form
Added message notifications in settings page

### DIFF
--- a/UserDetails/forms.py
+++ b/UserDetails/forms.py
@@ -29,8 +29,10 @@ class RegisterUserForm(UserCreationForm):
         fields = ('username', 'password1', 'password2', 'email')
 
     def clean(self):
+        cleaned_data = super(RegisterUserForm, self).clean()
+
         # Check if the email is not already used.
-        email = self.cleaned_data.get('email')
+        email = cleaned_data.get('email')
         if User.objects.filter(email=email).exists():
             msg = 'E-mail is already used.'
             self._errors['email'] = self.error_class([msg])
@@ -40,6 +42,7 @@ class RegisterUserForm(UserCreationForm):
 
 class RegisterUserDetails(forms.ModelForm):
     first_name = forms.CharField(max_length=40, required=True)
+    last_name = forms.CharField(max_length=40, required=True)
     allergies = forms.CharField(max_length=100, required=False, help_text="Max 100 characters, leave empty if none")
 
     class Meta:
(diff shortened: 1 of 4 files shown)


UserDetails/forms.py
```diff
@@ -29,8 +29,10 @@ class Meta:
         fields = ('username', 'password1', 'password2', 'email')
 
     def clean(self):
+        cleaned_data = super(RegisterUserForm, self).clean()
```
mhvis:
```suggestion
        cleaned_data = super().clean()
```

UserDetails/forms.py
```diff
@@ -29,8 +29,10 @@ class Meta:
...
+
         # Check if the email is not already used.
-        email = self.cleaned_data.get('email')
+        email = cleaned_data.get('email')
         if User.objects.filter(email=email).exists():
             msg = 'E-mail is already used.'
             self._errors['email'] = self.error_class([msg])
```
mhvis: I know this is not in the changeset, but I would use `raise ValidationError` or `add_error` here. It's not really good practice to modify private variables.

UserDetails/views_user.py
```diff
@@ -59,6 +60,8 @@ def post(self, request, *args, **kwargs):
         if self.context['form'].is_valid():
             self.context['form'].save()
             update_session_auth_hash(request, request.user)
+            messages.add_message(request, messages.ERROR, "Succesfully altered settings")
```
mhvis: `ERROR`?

UserDetails/views_user.py
```diff
@@ -81,6 +84,7 @@ def post(self, request, *args, **kwargs):
 
         if self.context['form'].is_valid():
             self.context['form'].save()
+            messages.add_message(request, messages.ERROR, "Succesfully altered settings")
```
mhvis: ^

assets/templates/accounts/register.html
```diff
@@ -31,6 +31,10 @@ <h3>Personal information</h3>
                         <h3>Memberships</h3>
                         {% include "snippets/bootstrap_form.html" with form=associationlink_form %}
 
+                        <br>
+                        <span class="text-danger">*</span> required<br>
```
mhvis: Use `<p></p>` instead of `<br>`, standalone text should almost always be put inside `p` tags, not doing so is considered bad practice.

UserDetails/views_user.py
DutcherNL: Oeps, die heb ik gemist XD

PR #88, opened 2019-03-06
Test cases

mhvis:
- Fixes users were able to join full dining lists.
- Fixes users were able to join closed dining lists.
- Tests updated.

The test cases are very minimal so interesting that they still discovered 2 bugs

### DIFF
--- a/Dining/models.py
+++ b/Dining/models.py
@@ -242,7 +242,7 @@ class DiningEntry(models.Model):
                 raise ValidationError(gettext('This user is already subscribed to the dining list.'))
 
             # (Optionally) validate if user is not already on another dining list
-            #if DiningList.objects.filter(date=self.dining_list.date, dining_entries__user=self.user)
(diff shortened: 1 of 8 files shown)


Dining/models.py
```diff
@@ -192,6 +192,14 @@ def external_dining_entries(self):
         """All dining entries that are not for external people."""
         return DiningEntryExternal.objects.filter(dining_list=self)
 
+    def can_modify(self, user):
+        """Users can modify when the dining list is open, owner can modify when dining list is still adjustable."""
```
DutcherNL: Should we not also take the purchaser into account here?

DutcherNL:
De view blokkeerde geloof ik de toegang in de implementatie. Dus alhoewel de form hier niet op staakte (goed gevonden trouwens) was het alsnog neit mogelijk op de eetlijst te komen.

Dining/models.py
mhvis: Oh yes thanks

PR #89, opened 2019-03-06
Settings page layout

DutcherNL:
Reimplemented tabular layout on settings page.
Restructured settings page layout
Altered Association membership forms


assets/templates/account/settings/settings_authorisation.html
```diff
@@ -0,0 +1,11 @@
+{% extends 'account/settings/settings_base.html' %}
+
+{% block tab_authorisation %}active{% endblock %}
+
+
+{% block details %}
+
+    See password_change.html for code
+    This is in a seperate file for allauth
```
mhvis: Why?

assets/templates/account/settings/settings_base.html
```diff
@@ -0,0 +1,30 @@
...
+        <li class="nav-item">
+            <a class="nav-link {% block tab_mail %}{% endblock %}"
+               href="{% url 'account_email' %}">
+                Mailing
+            </a>
+        </li>
+        <li class="nav-item">
+            <a class="nav-link {% block tab_authorisation %}{% endblock %}"
+               href="{% url 'account_change_password' %}">
+                Authorisation
```
mhvis: Please see definition of authorization vs authentication: https://en.wikipedia.org/wiki/Authentication#Authorization. But still, 'change password' is way more simple and direct.

assets/templates/account/settings/settings_base.html
```diff
@@ -0,0 +1,30 @@
+{% extends 'base.html' %}
+
```
mhvis: `settings` is double in the filename as the folder is already called `settings`.

So `account/settings/settings_base.html` should be `account/settings/base.html`, etcetera.

assets/templates/account/password_change.html
```diff
@@ -1,14 +1,11 @@
-{% extends "base.html" %}
+{% extends 'account/settings/settings_authorisation.html' %}
```
mhvis: Why not `account/settings/settings_base.html`???

assets/templates/account/password_change.html
```diff
@@ -1,14 +1,11 @@
...
-{% block content %}
-    <p><a href="{% url 'account_profile' %}">&larr; Back to profile</a></p>
-    <h1>{% trans "Change password" %}</h1>
-
+{% block details %}
     <form method="post" action="{% url 'account_change_password' %}">
         {% csrf_token %}
         {% include "snippets/bootstrap_form.html" %}
-        <button type="submit" name="action" class="btn btn-primary">{% trans "Change password" %}</button>
+        <p><button type="submit" name="user_submit" class="btn btn-primary btn-block">Save</button></p>
```
mhvis: Why are you changing the name attribute? I don't see that you use it elsewhere, so no reason to. You might confuse allauth.

UserDetails/views_user_settings.py
```diff
@@ -0,0 +1,41 @@
+from django.contrib import messages
+from django.contrib.auth.mixins import LoginRequiredMixin
+from django.shortcuts import render, redirect
+from django.views.generic import TemplateView
+from django.utils.translation import gettext as _
+
+from .forms import UserForm, DiningProfileForm, AssociationLinkForm
+
+
+class Settings_Profile_View(LoginRequiredMixin, TemplateView):
```
mhvis: Please use the Python naming conventions.

UserDetails/urls.py
```diff
@@ -15,10 +17,16 @@
...
+    path('statistics/', include([
+        path('dining/', DiningHistoryView.as_view(), name='history_lists'),
+        path('dining/<int:page>/', DiningHistoryView.as_view(), name='history_lists'),
+    ])),
+
 
-    path('history/dining/', DiningHistoryView.as_view(), name='history_lists'),
-    path('history/dining/<int:page>/', DiningHistoryView.as_view(), name='history_lists'),
-    path('profile/', ProfileView.as_view(), name='account_profile'),
+    path('settings/account', Settings_Profile_View.as_view(), name='settings_account'),
```
mhvis: `settings/account` -> `settings/account/`

UserDetails/urls.py
```diff
@@ -15,10 +17,16 @@
...
+    path('settings/', include('allauth.urls')),
+
+    # Override allauth login and sign up page with our registration page
+    path('login/', LoginView.as_view(), name='login_view'),
```
mhvis: Please use `account_login` as name for the login view as that is the same name that allauth uses.
```suggestion
    path('login/', LoginView.as_view(), name='account_login'),
```

mhvis:
These comments were some things that I noticed directly, I'll do a full review later.

UserDetails/forms.py
```diff
@@ -52,12 +57,141 @@ class Meta:
...
+            raise ValidationError("At least one association needs to be chosen")
+
+        return cleaned_data
+
+    def save(self, user=None):
+        """
+        Saves the association links. Removes
+        :return:
+        """
+        if self.user is None and user is None:
```
mhvis:
```suggestion
        if not self.user and not user:
```
I'd suggest for readability to only use `is None` when it is really needed. (Just a suggestion for the future)

UserDetails/forms.py
```diff
@@ -52,12 +57,141 @@ class Meta:
...
+            raise ValueError("Both self.user and user are None")
+        if user is None:
+            user = self.user
+
+        for key, value in self.cleaned_data.items():
+            link = self.fields[key].get_membership_model(user, new_value=value)
+            if value:
+                if link.id is None:
+                    link.save()
+                elif link.get_verified_state() == False:
```
mhvis: I think in Python `None == False`, so if you want to explicitly check for values that are `False` and not `None`, you need:
```python
elif link.get_verified_state() is False:
```

UserDetails/forms.py
```diff
@@ -52,12 +57,141 @@ class Meta:
...
+                    # If user was rejected, and a new request is entered
+                    link.verified_on = None
+                    link.save()
+            else:
+                if link is not None and link.get_verified_state() != False:
```
mhvis: See comment above, you probably want `is not` instead of `!=`. Also, if you here do not allow removal of rejected user memberships, I don't see the need of a cooldown period at all.

UserDetails/views.py
```diff
@@ -25,14 +25,14 @@ def get_context_data(self, **kwargs):
         context.update({
             'account_form': RegisterUserForm(),
             'account_detail_form': RegisterUserDetails(),
-            'associationlink_form': RegisterAssociationLinks(),
```
mhvis: `RegisterAssociationLinks` has no usages according to my PyCharm so it should be removed if you're not using it.

assets/templates/account/signup.html
```diff
@@ -28,15 +28,15 @@ <h3>Personal information</h3>
...
+                    {% include "account/snippet_associations_form.html" with form=associationlink_form %}
 
                     <p>
                         <span class="text-danger">*</span> required
                     </p>
 
                     <h3>Terms of service</h3>
                     {# Todo: privacy documents #}
-                    <div class="offset-sm-3 col-sm-8">
+                    <div class="offset-sm-2 col-sm-10">
```
mhvis: Sorry :laughing: I changed the Bootstrap form template to `sm-2`/`sm-10` when I saw that Bootstrap itself uses that.

assets/templates/account/snippet_associations_form.html
```diff
@@ -0,0 +1,63 @@
...
+            {% endif %}
+        </label>
+        {# Field label #}
+        <label for="{{ field.id_for_label }}" class="col-10">
+            <span class="{% if field.field.disabled %} text-muted {% endif %}">
+                {{ field.label }}
+            </span>
+
+            {% if field.field.verified %}
+                <span class="text-success">verified</span>
```
mhvis:
```suggestion
                <span class="text-success">Verified</span>
```
Also, badges might be nicer here.

UserDetails/forms.py
```diff
@@ -52,12 +57,141 @@ class Meta:
...
+                    if self.membership.is_verified:
+                        if self.membership.verified_on + \
+                                settings.DURATION_AFTER_MEMBERSHIP_CONFIRMATION > timezone.now():
+                            # The user has been verified to recently (prevent spamming)
+                            self.disabled = True
+                    else:
+                        if self.membership.verified_on + \
+                                settings.DURATION_AFTER_MEMBERSHIP_REJECTION > timezone.now():
+                            # The user has been verified not to be a member to recently (prevent spamming)
+                            self.disabled = True
```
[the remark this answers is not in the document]
DutcherNL: We can discuss this and always set it to 0 if necessary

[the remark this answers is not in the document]
DutcherNL: It was originally 1 cooldown period for both, but I figured that you might want to differ this, which is why I made it into two seperate parameters

UserDetails/forms.py
```diff
@@ -52,12 +57,141 @@ class Meta:
...
+
+        self.user = user
+
+        if user is None:
+            associations = Association.objects.filter(is_choosable=True)
+        else:
+            associations = Association.objects.filter(
+                Q(is_choosable=True) |
+                (Q(is_choosable=False) & Q(usermembership__related_user=user))) \
+                .distinct().order_by('slug')
```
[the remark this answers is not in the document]
DutcherNL: I don't think so either, but I do know that it originally was added because it was needed, though I may have altered the exact query contents since then. I'll have to double check this.

UserDetails/forms.py
```diff
@@ -52,12 +57,141 @@ class Meta:
...
+            field = AssociationLinkField(user, association)
+
+            self.fields[field.label] = field
+
+    def clean(self):
+        cleaned_data = super(AssociationLinkForm, self).clean()
+        has_association = False
+        for key, value in self.cleaned_data.items():
+            if value:
+                has_association = True
```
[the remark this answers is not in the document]
DutcherNL: Yup. That works too.

UserDetails/forms.py
DutcherNL: Ever since the querie code failed due to an incorrect None check on a queryset check I've been more rigid in my None checks, but I'll try to balance it out more.

UserDetails/forms.py
DutcherNL: I checked it, None == False returns False.
Also, in this particular instance, the effect would not matter as None indicates that the object has not been verified yet, which means that verified_on is already None. Setting it to None again doesn't change anything. The only effect is that it unnecessarily executes a database entry update while not changing any data.

UserDetails/forms.py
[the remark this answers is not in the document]
mhvis: Yup they should be different indeed and probably 0 for confirmed requests.

UserDetails/forms.py
DutcherNL: Ok. What happens in my approach is that if a member is denied and is still within the denial period the field becomes disabled. As a result the contents of this field aren't send along in a post request and can therefore not end up in this particular piece of code (although I don't know if this also occurs if you create a custom post request now that I think of it).
If a rejected membership ends here it can either be as False (i.e. it's not a membership) or True (the user wants to become a member).
- False: nothing should occur with the membership, otherwise you remove the rejection completely even if the user doesn't want to be a member
- True: The user was originally rejected, but still wants (/has become) a member and wants to request it again.
This particular piece of code deals with that the False case (186: if value:) where the membership should not be removed. The lines above 189-182. Takes care of the true case. The member was rejected, but wants a new request.
That said, I've shortened if link is not None to if link and... != False remains the same, as None != False results in True, which is what I want (along with True != False yielding True). I know it is a bit unorthodox.

UserDetails/forms.py
mhvis: Okay! I saw indeed that it in this case only leads to unnecessary db executions. Interesting that `None!=False` in Python, since `bool(None)==False`, thus it apparently does some other type conversion.

UserDetails/views.py
DutcherNL: It is removed? That's what this change is all about...

UserDetails/views_user_settings.py
DutcherNL: Oops, sorry. Renamed

assets/templates/account/password_change.html
DutcherNL: Reverted. Added btn-block in class

assets/templates/account/settings/settings_authorisation.html
DutcherNL: Allauth has views which you implemented, I didn't want to break on that with all custom views, however, to maintain some structure in the settings pages I started grouping them in a seperate folder which meant there was a disjoint in what to place where. This made, imo, the clearest solution.
"Why not account/settings/settings_base.html???"
As stated, I wanted all settings display logic in the settings folder, this includes the bit of code that autoselects the right tab which is in 'account/settings/settings_authorisation.html'

assets/templates/account/settings/settings_base.html
DutcherNL: what does a shorter name save? This way you can see a single html file and have some idea what it is intended for (with a file named base.html I assume it's the root of the project, not the root of a small part on a single template-chain). I also like template chains not call two different files with the same name if possible.

assets/templates/account/settings/settings_base.html
DutcherNL: Yes, yes I did. It's now authentication in naming and the tab is called password.
The reason I wanted authentication is because we can also place Q-link relevant information on this page in the future

assets/templates/account/snippet_associations_form.html
DutcherNL: 👍

assets/templates/account/settings/settings_base.html
mhvis: It still says Authorization for me.

assets/templates/account/settings/settings_base.html
mhvis: It's clearer for reading, the path is also considered to be part of the file, you'll often see it how I described

assets/templates/account/settings/settings_authorisation.html
mhvis: All templates in the `account` folder are ours and should not be considered as part of allauth, although I get that that's probably not what you meant. This extra separation is very verbose in this case because they are already very close and therefore that does not weigh up to the extra files that are needed to make it work. The extra files make that more work is needed to understand what's going on (more files to follow).

UserDetails/views.py
mhvis: I mean the implementation in `UserDetails/forms.py:37`. According to PyCharm it is not in use anywhere.

assets/templates/account/snippet_associations_form.html
mhvis: It's still `verified` instead of `Verified`

assets/templates/account/settings/settings_authorisation.html
DutcherNL: It is indeed not what I meant. What I meant was that the allauth views reference those files and as we did not customise those views, the exact location of those files can not be changed either.

UserDetails/views.py
DutcherNL: O, de class zat er nog in. Ah. Zat. zat er in. Zat.

assets/templates/account/settings/settings_authorisation.html
mhvis: You can still just directly extend the base template, the extra indirection does not really help for clearability or maintainability.

UserDetails/views.py
mhvis: :sunglasses:

PR #104, opened 2019-04-22
Allauth Quadrivium

### LINKED ISSUE #90: Quadrivium login does not use allauth library

mhvis:
Fixes #90 and makes external logins more modular and easier to extend with other account providers. Also removes dependency of the separate OpenID Connect library.


UserDetails/views_user_settings.py
```diff
@@ -29,13 +29,13 @@ def post(self, request, *args, **kwargs):
         })
 
         if context['user_form'].is_valid() and \
-            context['dining_form'].is_valid() and \
-            context['association_links_form'].is_valid():
+                context['dining_form'].is_valid() and \
```
DutcherNL: Why the increment?

assets/templates/socialaccount/connections.html
```diff
@@ -0,0 +1,47 @@
...
+            {% endfor %}
+
+            <div>
+                <button type="submit">{% trans 'Remove' %}</button>
+            </div>
+
+        </form>
+
+    {% else %}
+        <p>{% trans 'You currently have no social network accounts connected to this account.' %}</p>
```
DutcherNL: I'd prefer external accounts over social accounts. Social implies Facebook and Twitter and that's not what we do or currently aim for here.

mhvis:
Thank's for the review already :smile:, it's still a draft and not finished so I'll let you know when it's finished.

UserDetails/views_user_settings.py
mhvis: I'm just using the built-in PyCharm code formatter here. In this case it's obviously done to have a visual indication of the if statement (consisting of 3 lines) and the first statement after the if on the 4th line.

mhvis:
I have now the following for the log in page. The linked accounts page is also a bit similar.

mhvis:
Automatic membership creation is now added and users that sign up using external account are redirected to the settings page so they directly get the option to set allergies/dietary wishes. It's now reasonably finished but I'll still have to test it. There will probably not be much changes anymore so you can already see if you're okay to merge. After merging the already existing accounts on the server need to be updated manually, so I'll want to merge this one myself :smile_cat:. After that I'll remove the old code and dependency.

mhvis:
It's now finished and works much better than previously so I'd like to merge if you're okay @DutcherNL :smiley:

PR #105, opened 2019-04-22
Margin fix

mhvis:
I came across a margin bug on this page where there is not side margin next to the tabs and heading.

That was due to not having content placed inside of a `col`, instead merged with it :smirk: 

```html
<div class="row">
    <h2 class="col-12 pl-0">{{ association }}</h2>
</div>
```

In that case, the padding on the col that undoes the margin on the row is not applied which you can see using the dev console:

I've obviously fixed it here by removing the grid layout but if you want to keep it you can change it yourself to be like this :yum:

```html
<div class="row">
    <div class="col-12"><h2>{{ association }}</h2></div>
</div>
```

This is the end result:

### DIFF
--- a/assets/templates/accounts/associations_base.html
+++ b/assets/templates/accounts/associations_base.html
@@ -1,31 +1,23 @@
 {% extends 'base.html' %}
 
 {% block content %}
-        <div class="row">
-            <h2 class="col-12 pl-0">{{ association }}</h2>
-        </div>
-        <div class="row">
-            <ul class="col-12 nav nav-tabs">
-                <li class="nav-item">
-                    <a class="nav-link{% block tab_overview %}{% endblock %}"
-                       href="{% url 'association_overview' association_name=association.slug %}">Overview
-                        {% if notify_overview %}<span class="badge badge-pill badge-warning">!</span>{% endif %}
-                    </a>
-                </li>
-                <li class="nav-item">
-                    <a class="nav-link{% block tab_credits %}{% endblock %}"
-                       href="{% url 'association_credits' association_name=association.slug %}">Credits</a>
-                </li>
-                <li class="nav-item">
-                    <a class="nav-link{% block tab_members %}{% endblock %}"
-                       href="{% url 'association_members' association_name=association.slug %}">Members</a>
-                </li>
-            </ul>
-        </div>
+    <h2>{{ association }}</h2>
+    <ul class="nav nav-tabs">
+        <li class="nav-item">
+            <a class="nav-link{% block tab_overview %}{% endblock %}"
+               href="{% url 'association_overview' association_name=association.slug %}">Overview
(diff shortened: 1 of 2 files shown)


mhvis:
I overlooked this earlier somehow but the issue is also on the History page, can fix it tomorrow

DutcherNL:
I actually did that deliberately as it is basically a header to the page creating a clearer difference between the tabs and the contents of the tabs. It was a styling choice for mobile format.

mhvis:
In that case you should've created an explicit solution instead of using a hack for that. That way it can be consistently applied everywhere. An example explicit solution would be to create this custom class:

```css
.container-extended {
    margin-left: -10px;
    margin-right: -10px;
}
```

This way you still have 5px space leftover which is already nicer than having it touch the viewport, but still those details are so small and meaningless because everyone has its own opinion on whether she likes it more when it's extended or when not. Apparently the Bootstrap team likes it better when it's not extended to the viewport, otherwise they would have added something for that. Taking into account that some people like the one and others like it differently, I'd recommend to always go for the simplest solution that requires the least amount of tags and don't worry too much about small details like those. People are not going to judge you for that the tabs do not extend to the viewport.

DutcherNL:
For clarity I try not to make any more new css classes in new files and rely as much on Bootstrap as possible so bootstrap files can be exchanged more easily without conflicting in other css files (and yes, I know we still use some unique css classes for the login page and the dining page), but I've changed and try to minimize it.

mhvis:
There's no problem in using custom classes to modify default behavior of a CSS framework. It's usually needed or desired to modify a certain CSS framework to add website specific styles. There are 2 ways that I've seen:

* Using a custom build pipeline (often using webpack) so that SASS can be used which will be compiled into a single `bundle.css` file which has Bootstrap and all modifications. That's really cool since you can modify Bootstrap variables like the primary color. Disadvantage is obviously that you need another build pipeline. Here's an example of it, see the files `_custom.scss` and `_variables.scss`: https://github.com/mhvis/klassiek-op-de-campus/tree/master/assets/style
* Using two stylesheets. Then you'd first include the Bootstrap stylesheet on the page and after that you include a stylesheet that adds or modifies styles (later includes override earlier includes). It was usually better to only have 1 extra stylesheet which has all custom styles since that reduces the number of needed requests, but with the new HTTP 2.0 there is no request limit anymore so that doesn't really matter anymore. These stylesheets are normally only downloaded once and subsequent page visits use a cached version (if set up correctly). That's most easiest for us since we'd only have one custom stylesheet.

So shall I merge the stylesheets that we have now and create a single `style.css`?

mhvis:
Coming back to this PR, the tab layout on this page is currently in `master` still inconsistent with the tabs on the account page, and also not conforming to Bootstrap, so I don't see why this PR would be problematic to merge as it fixes both. I can improve the stylesheet layout here directly as well if you want.

PR #106, opened 2019-04-22
GitHub link on Report Bug page

mhvis:
Suggestion for a simple link to the repository. Also cleaned the images folder in this branch.


DutcherNL:
I agree with all except for the removal of the following files:
- shopping cart.png
- dishes.png
- cooking.png
I want to use these for the cooking, shopping and cleaning stats at some point. Awesomefont doesn't have good icons for those three stats. I'm probably gonna convert them to icons, though I haven't researched yet how. Untill then I'd like to have these files as back-up.

mhvis:
They're back now

PR #109, opened 2019-04-23
Increase dish max_length upon request

mhvis:


DutcherNL:
It's probably gonna mess up styling a bit when we have superlong dish names, but we can always revert it :)

PR #116, opened 2019-04-25
Maximum debt limit exception for associations

### LINKED ISSUE #108: Q wants larger negative limits for their users
Q uses an incasso service within their association. They want to apply the dining costs to this incasso service as well which would mean that their users can rank up larger debts in the system

gwvdh:
Added to the settings the possibility to add associations to not have a maximum debt for their members.
Members which are a member of any those association will be able to join dinner lists even if they exceed maximum debt.

Closes #108


ScalaApp/scala_settings.py
```diff
@@ -23,6 +23,9 @@
 MINIMUM_BALANCE_FOR_DINING_SLOT_CLAIM = Decimal('-2.00') - KITCHEN_COST
 MINIMUM_BALANCE_FOR_USER_TRANSACTION = Decimal('0.00')
 
+# Except associations from bottom limit by association id
+MINIMUM_ASSOCIATION_EXCEPTIONS = []
```
DutcherNL: For optimal use this should be softcoded not hardcoded. I.e. This is a setting that should be in the Association model stored on the database not somewhere in the code.

UserDetails/models.py
```diff
@@ -74,6 +74,21 @@ def has_any_perm(self):
...
+
+    def is_member_of(self, associationId):
+        '''
+        Return if the user is a member of the association identified by its id
+        '''
+        for m in UserMembership.objects.filter(related_user=self):
+            if (m.association.id == associationId and m.is_verified):
+                return True
+        return False
+
```
DutcherNL: I like this change in user checks, thank you for that

ScalaApp/scala_settings.py
mhvis: Do you want it that way? It's probably not a change that will be made lightly and it can be compared with e.g. the kitchen cost setting which is also hardcoded. It's a setting which means that it's meant to be changeable for each deployment.

mhvis:
This closes #108

ScalaApp/scala_settings.py
DutcherNL: Changing a limit is indeed not something that should be taken lightly, however, it should also not be something that should potentially need to go through the ICT commission in my opinion. If Q wants an unlimited balance, I'm gonna assume other associations also want some custom arrangement (possibly with higher, but not unlimited depths, though I still need to inquire that). Thus any adjustment should be visible in the system itself instead of in the code and should be able to be adjusted by the Scala board directly if neccessary.
Also, I personally think it is weird to hardcorde an exception for a specific instance in the database table.

gwvdh: Now added as a tickbox in the admin panel

mhvis:
Heb geupdate op server en Q op unlimited gezet

PR #120, opened 2019-04-26
Date navigation relayout

### LINKED ISSUE #110: UI missing for navigating to days outside current week
It's not possible via the interface to go to days which are more than a week in advance or to go to past days.

### LINKED ISSUE #119: Association pictures sometimes flash during loading
Improvements:

* Reduce the image sizes, they're now huge (around 2000x1000 pixels each)
* Use `img` tag instead of CSS, that way they're loaded immediately and do not have to wait for the processing of CSS to start to begin loading.
* The images are currently not fully cached, they return the `Cache-Control: max-age=0` header. This should be improved in the nginx server settings.

mhvis:
Suggestion for fix #110

### DIFF
--- a/Dining/datesequence.py
+++ b/Dining/datesequence.py
@@ -1,0 +1,56 @@
+from datetime import date, timedelta
+
+
+class BaseSequencedDate(date):
+    """A date that is in a sequence of dates and has next/previous methods to move. This base class is a sequence
+    over all dates. To create different sequences, subclass this and override the upcoming method."""
+
+    @classmethod
+    def upcoming(cls, from_date=None, reverse=False):
+        """Returns the closest date in the sequence after given date, or before given date if reverse is True.
+        If given date is in the sequence, this date itself is returned! (But as a sequence subclass.)
+        If from_date is None, current date is used."""
+        if not from_date:
+            # Might want to use django.utils.timezone if we're going global
+            return cls.today()
+        return cls(from_date.year, from_date.month, from_date.day)
+
+    def next(self):
+        return self.upcoming(self + timedelta(days=1))
+
+    def previous(self):
+        return self.upcoming(self - timedelta(days=1), reverse=True)
+
+    @classmethod
+    def in_sequence(cls, d):
+        """Returns True if the given date is in the sequence of this class, False otherwise"""
+        return cls.upcoming(d) == d
+
+    @classmethod
+    def fromdate(cls, d):
+        """Creates an instance using any date instance, raises ValueError when given date is not in the sequence"""
+        if not cls.in_sequence(d):
+            raise ValueError("Date is not in the sequence")
+        return cls(d.year, d.month, d.day)
+
+
+class WeekdaySequencedDate(BaseSequencedDate):
+    """Sequence consisting of all weekdays"""
+
+    @classmethod
+    def upcoming(cls, from_date=None, reverse=False):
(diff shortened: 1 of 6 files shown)


mhvis:
I've generalized the date sequence so that we can easily change the dates so that e.g. the summer holiday is excluded from available dates (see the code). This is the updated layout:









During weekends (imagine 'Monday' instead of 'Tuesday'):

Dining/datesequence.py
```diff
@@ -0,0 +1,56 @@
...
+                # to next Monday
+                d = d + timedelta(days=7 - d.weekday())
+            else:
+                # to previous Friday
+                d = d - timedelta(days=d.weekday() - 4)
+        return super().upcoming(d)
+
+
+# Date sequence class that is in use in the application. Can change this into a Django setting to have it
+# deployment specific (could e.g. exclude summer holiday)
```
DutcherNL: Just a note in case you didn't know: we also have the kitchen during all holidays

assets/templates/dining_lists/dining_day.html
```diff
@@ -1,37 +1,65 @@
...
+                                <i class="fas fa-chevron-left"></i>
+                                <span class="sr-only">Previous</span>
+                            </a>
+                        {% endwith %}
+                        {% with d=date.next %}
+                            <a href="{% url 'day_view' year=d.year month=d.month day=d.day %}"
+                               class="btn btn-outline-primary">
+                                <i class="fas fa-chevron-right"></i>
+                                <span class="sr-only">Next</span>
+                            </a>
```
DutcherNL: In the beta trial, someone resereved a slot for 10 years in the future. Thus the reservations of slots has been capped to 30 days to prevent any possible future conflicts. How do we want to represent that in the UI?

assets/templates/dining_lists/dining_day.html
```diff
@@ -1,37 +1,65 @@
...
+                    {% endwith %}
+                </div>
+                <div class="col-7">
+                    <div class="btn-group btn-block">
+                        {% with d=date.previous %}
+                            <a href="{% url 'day_view' day=d.day month=d.month year=d.year %}"
+                               class="btn btn-outline-primary">
+                                <i class="fas fa-chevron-left"></i>
+                                <span class="sr-only">Previous</span>
+                            </a>
```
DutcherNL: Currently dininglist transactions are pending for two days after which entries  should no longer be removed or added (because it won't represent the cost). To prevent abuse of the helping statistics a similar limit might also be needed on the stats (to prevent dining list owners tweaking their own entries on very old dininglists. Do we want to represent the unchangable limits to the UI (possibly limit going backwards for more than two days)?

mhvis: Don't see any need for indicating it in the UI, if you want we can show a similar warning. If you want to hide the previous button we should instead block access altogether to these dates. We could do that by making the date sequence bounded

assets/templates/dining_lists/dining_day.html
mhvis: We could also block access to these days if nothing can happen there anyway (like other comment). That would be my preferable solution for preventing creation of dining slots, I'd be fine with that. I don't really think it's needed though, I expect the users to be good citizens. People could also just claim dining spots and not cook and we also expect them to not do that. If people do these things, so be it, and if it gets really annoying then we'll take care of it

PR #122, opened 2019-04-29
Dining list membership display

DutcherNL:
Memberships are now displayed on the dining list.
Dining list entries ordered by first name


UserDetails/migrations/0013_merge_20190429_1911.py
```diff
@@ -0,0 +1,14 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('UserDetails', '0012_auto_20190429_1306'),
+        ('UserDetails', '0011_auto_20190429_1748'),
+    ]
+
+    operations = [
+    ]
```
mhvis: These 2 migrations are probably going to be problematic because migrations up to 12 are already applied on the server.  Better to take the migrations from `master` and then create a new migration.

assets/templates/snippets/snippet_userdisplay.html
```diff
@@ -0,0 +1,17 @@
+{% load static %}
+
+<span class="d-flex justify-content-between align-items-center">
+    <span>{{ entry.name }}</span>
+
+    <span class="d-flex align-items-right">
```
mhvis: `align-items-right` is not an existing css class as far as I know. You can remove both classes here I think, just leave `<span>` without the class argument

assets/templates/snippets/snippet_userdisplay.html
```diff
@@ -0,0 +1,17 @@
...
+        {% for membership in entry.user.get_memberships %}
+            {% if membership.association.icon_image %}
+                <i class="membership_icon">
+                    <img src="{{ membership.association.icon_image.url }}">
+                </i>
+            {% else %}
+                <i>{{ membership.association.slug|slice:":1" }}</i>
```
mhvis: Neat

UserDetails/migrations/0013_merge_20190429_1911.py
DutcherNL: I made a migration file before I had a local copy of the migration files from the Wessels branch. The last migration file is a merge file, so it should not have a conflict. But I changed it anyhow.

assets/templates/snippets/snippet_userdisplay.html
DutcherNL: Nope, it wasn't. align-items-center was, so I thought align-items-right did as well. Funny enough, removing align-items-center resulted in a default behaviour that I wanted, so I didn't notice it XD

PR #124, opened 2019-04-30
Added csv view to get all diners and memberships info

DutcherNL:
I need a csv overview where I know which association diners are from and how often they have eaten. It is required for the left-over cost division key made back in 2018 (with deadline may 1st


DutcherNL:
It is speed temporary, however I expect the data to be used in future division keys as well. So it will likely be permanent later on (i.e. next year) and yes, I have elaborately tested it.

It actually does include external entries as well. Recall that DiningEntry is a superclass for DiningEntryInternal and DiningEntryExternal.

PR #126, opened 2019-05-02
Chef authorisation fix

DutcherNL:
Implemented option for dining owner to add diners over the maximum level
Allowed purchaser to also add people past the sign up deadlines as well as max limits


Dining/models.py
```diff
@@ -271,8 +273,11 @@ class DiningEntryUser(DiningEntry, DiningWork):
     def clean(self):
         super().clean()
         if not self.pk:
-            if DiningEntryUser.objects.filter(dining_list=self.dining_list, user=self.user):
-                raise ValidationError(_("User is already on this dining list."))
+            try:
+                if DiningEntryUser.objects.filter(dining_list=self.dining_list, user=self.user):
```
mhvis: Can use `exists`: https://docs.djangoproject.com/en/2.2/ref/models/querysets/#django.db.models.query.QuerySet.exists

Dining/models.py
```diff
@@ -271,8 +273,11 @@ class DiningEntryUser(DiningEntry, DiningWork):
...
+                    raise ValidationError(_("User is already on this dining list."))
+            except ObjectDoesNotExist:
```
mhvis: Where can this exception be raised in the try block?

Dining/templatetags/dining_tags.py
```diff
@@ -1,5 +1,26 @@
 from django import template
 
+from Dining.models import DiningEntry
+
 register = template.Library()
 
-# Todo: this file is currently not used and can maybe be deleted
+
+@register.filter(name='can_join_slot')
```
mhvis:
```suggestion
@register.filter
```

`name` argument is not needed, defaults to name of function

ScalaApp/scala_settings.py
```diff
@@ -19,8 +19,8 @@
 KITCHEN_USE_END_TIME = time(19, 30)
 
 # Balance bottom limit
-MINIMUM_BALANCE_FOR_DINING_SIGN_UP = Decimal('-2.00') - KITCHEN_COST
-MINIMUM_BALANCE_FOR_DINING_SLOT_CLAIM = Decimal('-2.00') - KITCHEN_COST
+MINIMUM_BALANCE_FOR_DINING_SIGN_UP = Decimal('-2.00') + KITCHEN_COST
+MINIMUM_BALANCE_FOR_DINING_SLOT_CLAIM = Decimal('-2.00') + KITCHEN_COST
```
mhvis: :smile: Wessel noticed it but we forgot to tell you this

Dining/forms.py
```diff
@@ -278,17 +277,17 @@ def __init__(self, deleted_by, instance, **kwargs):
...
         # Check permission
-        if self.deleted_by != self.instance.user and self.deleted_by != list.claimed_by:
+        if self.deleted_by != self.instance.user and not dining_list.is_authorised_user(self.deleted_by):
             raise PermissionDenied('Can only delete own entries')
 
         # Validate dining list is still open (except for claimant)
-        if not list.is_open():
-            if self.deleted_by != list.claimed_by:
+        if not dining_list.is_open():
+            if dining_list.is_authorised_user(self.deleted_by):
```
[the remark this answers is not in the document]
DutcherNL: Test case added

Dining/models.py
```diff
@@ -107,6 +107,9 @@ def get_purchaser(self):
         else:
             return self.purchaser
 
+    def is_authorised_user(self, user):
+        return user == self.claimed_by or self.purchaser
```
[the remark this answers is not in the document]
DutcherNL: omg, wat een slippertje

Dining/models.py
```diff
@@ -147,17 +150,21 @@ def has_room(self):
         Determines whether this dining list can have more entries
```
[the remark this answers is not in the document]
DutcherNL: I personally want the adjustable period to apply to the admin panel as well, though not that strongly. We'll talk about it later.

Dining/models.py
DutcherNL: Due to the clean_dining_list method in the create form it could happen that the cleaning of dining_list fails (list closed, full, whatever). As a result, the code calling clean_dining_list is returned an error instead of a value (setting self.dininglist to None). However, django still runs the models clean method afterwards for some reason. Thus in the filter query a query with dininglist=None is executed returning an objectDoesNotExist. I don't know of a better solution than to remove this clean method here entirely and moving it to the form instead (though arguably we don't want any case of a duplicate dining entry in the list).
The phrasing in the validationerror was incorrect though, that has been fixed.

ScalaApp/scala_settings.py
DutcherNL: And you didn't think to fix it yourself... boooo :P

mhvis:
The individual methods are still missing unit tests, I suggest in the future to go more towards a TDD (test driven development) workflow where you first add testcases for the new methods that you're adding, before you actually implement the methods themselves.

PR #132, opened 2019-05-06
EntryAddView

### LINKED ISSUE #36: User add has two steps
When adding a user to a dining list you search for the user and then select a user on the next page. This can be combined using an auto-complete field on which you directly select a user.

mhvis:
Fixes #36. I've opted for a client-side solution where all user full names are stored on the client on page load, as the number of users will not be so large that this becomes a problem. Scraping of the names of all users is possible in both approaches anyway.

Used library: https://github.com/leaverou/awesomplete/

There are also some changes to forms, the changes here are an example of the changes to forms that I want to apply everywhere, that is: get rid of the `__init__` methods and replace those with using `instance` on creation, get rid of fields that are not actually user input (`dining_list`) and move business logic from model clean methods to form clean methods.

Can't easily get a screenshot of the autocomplete list, but here's something:


mhvis:
I came across another error in your former PR (#126) which I fixed in the last commit. There are probably more, this part definitely needs a rewrite and testcases.

CreditManagement/management/commands/end_beta.py
```diff
@@ -1,12 +0,0 @@
...
-from Dining.models import DiningList, DiningEntry, DiningEntryExternal
-
-class Command(BaseCommand):
-    help = 'Finalises all expired transactions'
-
-    def handle(self, *args, **options):
-        FixedTransaction.objects.all().delete()
-        DiningList.objects.all().delete()
-        DiningEntry.objects.all().delete()
-        DiningEntryExternal.objects.all().delete()
```
DutcherNL: I completely forgot we still had this XD

PR #133, opened 2019-05-07
Feat search user for transaction

DutcherNL:
Added autosearch widget and implemented them on the transaction forms

### DIFF
--- a/CreditManagement/views.py
+++ b/CreditManagement/views.py
@@ -4,8 +4,6 @@ from django.views.generic.list import ListView
 from CreditManagement.models import *
 from django.contrib import messages
 from django.utils.translation import gettext as _
-from django.contrib.auth.decorators import login_required
-from django.utils.decorators import method_decorator
 from .forms import UserTransactionForm, AssociationTransactionForm
 from django.views.generic import View
 from django.http import HttpResponseForbidden, HttpResponseRedirect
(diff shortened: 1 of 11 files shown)


CreditManagement/views.py
```diff
@@ -24,7 +22,7 @@ class TransactionAddView(LoginRequiredMixin, View):
     template_name = "credit_management/transaction_add.html"
     context = {}
 
-    def get(self, request, association_name=None, *args, **kwargs):
+    def get(self, request, association_name=None):
```
mhvis: I've had some problems in instances where `*args` and `**kwargs` were left out for these request handling methods (get/post). If it works for you then fine, but maybe check it.

assets/templates/general/widget_search_select.html
```diff
@@ -0,0 +1,37 @@
...
+        // Enable autocomplete
+        new Awesomplete(searchfield, {
+            list: [{% for entry in widget.queryset %}["{{ entry }}","{{ entry.id }}"],{% endfor %}],
+            replace: function (text) {
+                this.input.value = text.label;
+            },
+            autoFirst: true,
+        });
+
+        // Update internal button when user selects someone
```
mhvis: Outdated comment

DutcherNL: Yup, changed it in last commit:
// Update the selected instance-value

PR #141, opened 2019-05-10
Dining entry delete rewrite

mhvis:
Continuation of #135. Minor functionality changes, I believe the only thing is that you can now delete a dining user entry from someone else (not external) if you created that one in the first place.

Edit: and minor payment form change

### DIFF
--- a/Dining/forms.py
+++ b/Dining/forms.py
@@ -110,54 +110,50 @@ class CreateSlotForm(ServeTimeCheckMixin, forms.ModelForm):
 
 
 class DiningInfoForm(ServeTimeCheckMixin, forms.ModelForm):
-    def __init__(self, *args, **kwargs):
-        dining_list = kwargs.get("instance")
-        super(DiningInfoForm, self).__init__(*args, **kwargs)
-
-        query = dining_list.diners.distinct()
-        self.fields['purchaser'].queryset = query
-
     class Meta:
         model = DiningList
         fields = ['serve_time', 'min_diners', 'max_diners', 'sign_up_deadline', 'purchaser']
 
-    def save(self):
-        self.instance.save(update_fields=self.Meta.fields)
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields['purchaser'].queryset = self.instance.diners.distinct()
 
 
 class DiningPaymentForm(forms.ModelForm):
-    dinner_cost_total = forms.DecimalField(decimal_places=2, max_digits=5, initial=Decimal(0.00),
-                                           validators=[MinValueValidator(Decimal('0.00'))])
+    dinner_cost_total = forms.DecimalField(decimal_places=2, max_digits=5, required=False,
+                                           validators=[MinValueValidator(Decimal('0'))],
+                                           help_text='Only one of dinner cost total or dinner cost per person should be provided')
 
     class Meta:
         model = DiningList
         fields = ['dish', 'dinner_cost_total', 'dining_cost', 'payment_link']
-        save_fields = ['dish', 'dining_cost', 'payment_link']
-        help_texts = {
-            'dinner_cost_total': 'Either adjust total dinner cost or single dinner cost',
-            'dinner_cost_single': 'Either adjust total dinner cost or single dinner cost',
-        }
 
-    def __init__(self, *args, **kwargs):
-        super(DiningPaymentForm, self).__init__(*args, **kwargs)
+    def clean(self):
+        """This cleaning calculates the person dining cost from the total dining cost"""
+        cleaned_data = super().clean()
+        dinner_cost_total = cleaned_data.get('dinner_cost_total')
+        dining_cost = cleaned_data.get('dining_cost')
 
-    def save(self):
-        # If the single value has been added, recompute the total amount
-        # also check if it has changed from earlier status
-        old_list = DiningList.objects.get(id=self.instance.id)
-
-        total_dinner_cost = self.cleaned_data['dinner_cost_total']
-
-        if total_dinner_cost > 0:
+        # Sanity check: do not allow both dinner cost total and dinner cost per person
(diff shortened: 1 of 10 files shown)


assets/templates/dining_lists/dining_slot_diners.html
```diff
@@ -6,176 +6,163 @@
...
+                <tbody>
+                {# Table rows #}
+                {% for entry in entries %}
+                    {% with entry=entry.get_subclass %}
+                        <tr>
+                            <td>
+                                <div class="d-flex justify-content-between align-items-center">
+                                    <span>
+                                        {{ entry.get_name }}
+                                        {% if entry.is_external or entry.created_by != entry.user %}
```
DutcherNL: I'm not a fan of displaying the user that added the account; it serves no purpose and can clutter the list (especially in mobile format). The reason it is displayed on external users is that the claimer can trace where the new person originates from. However, for already existing accounts this is less of an issue. Thus I'd like to see the added by no longer displayed on internal entries.

assets/templates/dining_lists/dining_slot_diners.html
```diff
@@ -6,176 +6,163 @@
...
-                                        </span>
+    <div class="row">
+        <div class="col-12 table-responsive">
+            <table class="table table-hover">
+                {# Table header #}
+                <thead>
+                <tr class="">
+                    <td style="min-width: 10rem">Name</td>
+                    <td style="min-width: 15rem;">Help stats</td>
+                    <td></td>
```
DutcherNL: If nothing can be deleted, there is no reason to reserve space for the column. And yes, I know it won't have content and thus be smaller, But running here locally still shows a total width of 13 interior + 2*12 padding. That's why I had the can_delete_some paramater. It was a conscious styling choice.

assets/templates/dining_lists/dining_slot_diners.html
```diff
@@ -6,176 +6,163 @@
...
+                {% url 'entry_add' day=date.day month=date.month year=date.year identifier=dining_list.association.slug %}">
+                   + Add Diner</a>
+           {% endif %}
 
-            </div>
         </div>
-    </form>
+    </div>
 
 {% endblock details %}
```
DutcherNL: I noticed that for some reason, on my desktop, the mouse doesn't hover to the hand/finger anymore when hovering the buttons. Any idea why?

Dining/forms.py
```diff
@@ -110,54 +113,50 @@ def save(self, commit=True):
...
-
-        self.instance.save(update_fields=DiningPaymentForm.Meta.save_fields)
+                raise ValidationError({
+                    'dinner_cost_total': "Can't calculate dinner cost per person as there are no diners"})
+
+            # Round up to remove missing cents
+            cost = Decimal(cost).quantize(Decimal('.01'), rounding=ROUND_UP)
+            cleaned_data.update({
+                'dinner_cost_total': None,
+                'dining_cost': cost,
```
DutcherNL: Good addition :)

Dining/views.py
```diff
@@ -484,57 +477,63 @@ def post(self, request, *args, **kwargs):
...
+        ## Story time
+        This suffered from the most awesome bug. Earlier, one form would be saved if it was valid, even when the other
+        form wasn't valid. E.g. when the info form was not valid while the payment form is actually valid, the payment
+        form would get saved. However the payment form uses the same dining list instance as for the info form, and
+        invalid field values of the info form do get applied to the dining list instance, they are just normally not
+        saved in the database because the form would be invalid. However the payment form is valid and therefore saves
+        the dining list anyway, with the invalid field values.
+        
+        An explicit include of only fields that were part of the form during saving prevented the bug from manifesting
+        (perhaps it was meant that way?).
```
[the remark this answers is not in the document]
mhvis: That a part of the form is still saved is indeed actually the only bug as seen from the user. Still the use of exclude as workaround is worse in my opinion

assets/templates/dining_lists/dining_slot_diners.html
mhvis: People are not going to be mad at you about that (they are actually not even going to notice)

assets/templates/dining_lists/dining_slot_diners.html
mhvis: I notice it indeed, only change is that it's not directly inside form tags anymore. I'll fix the pointer.

assets/templates/dining_lists/dining_slot_diners.html
mhvis: I'll change it

mhvis:
About your comment about the large overhaul of `dining_slot_diners.html`, if you look closely I think you'll notice that I actually didn't change a lot. A large diff doesn't immediately mean that there are a lot of changes, in this case the large diff is caused by that the indentation for the whole block is changed since the outer form tag is removed.

The changes in that file are (maybe I'm forgetting something):

* I removed the outer form tag since that one prevented that we could use forms inside the table for the delete buttons. Now that the delete buttons are forms directly, the JavaScript workaround is not needed anymore and thus removed.
* The delete column is always there since extra code would be needed to hide it, and extra code for some view thing that users are not even going to notice is not really worth it. (If it's possible using a single line, then okay)

These are I believe the only 2 changes

assets/templates/dining_lists/dining_slot_diners.html
mhvis: Actually, it's not a pointer (hand) currently on the live server as well, since it's a label tag and not a button which doesn't get a pointer on hover. I'll still change it.

mhvis: I'll make an issue instead and we'll fix it in a separate PR

assets/templates/dining_lists/dining_slot_diners.html
DutcherNL: No, but I am. That's the thing.. haha

PR #144, opened 2019-05-12
django-autocomplete-light

mhvis:
I have to bring over my condolences once again, I've once again been on a killing spree :gun:. This time my suggestion is to get rid of the Awesomplete library, after it's now in use for less than a week :laughing:

The thing is, while looking for the needed components for multiple owners, one thing that was needed was a widget for a many to many relationship combined with autocomplete, and while looking for that I came across this library. I guess I overlooked it previously. It's probably the much better choice over Awesomplete despite it's disadvantage which is that now the add diner page needs 2 fields and it's maybe uglier. The new library however has Django widgets for foreign key and many to many fields as well as the autocomplete uses server side queries with easy to use Django views and surely a lot more flexible. This seems to be the de facto library for this, should probably have gone for it directly. So this PR is mostly as a preparation for when we're going to need many to many widgets.

Sorry for the large number of commits, I made an error somewhere with rebasing.

~NOTE: there's 1 functional change here, which is that the create new transaction form for an association allows selecting all users instead of only association members! The queryset filter that filtered for the association is no longer there now as that would require a separate endpoint as it's server side now. It's still quite easily possible to restrict creating transactions to members only after submitting the form. However non members would still show up in the user selection box, as that uses the server-side user query view which does not filter for the association. After submitting a user that's not a member, an error would appear beneath the field. So here the question: are transactions to non association members allowed? I don't see the problem with that directly. If it's not allowed I have to update it.~


assets/templates/snippets/bootstrap_form.html
```diff
@@ -28,33 +28,41 @@
...
-            {# Field widget #}
-            {% if field.field.disabled %}
-                {# Readonly widget #}
-                {% render_field field class="form-control-plaintext" placeholder=field.label %}
-            {% elif form.is_bound %}
-                {# Validated widget #}
-                {% if field.errors %}
-                    {# Errors present #}
-                    {% render_field field class="form-control is-invalid" placeholder=field.label %}
+            {% if field|widget_type == "modelselect2" or field|widget_type == "modelselect2multiple" %}
```
DutcherNL: It's fine for now, but this is exactly what bothered me with the autocomplete2 widgets. Ideally you don't want to be conserned with the exact widget type in the template, but in order to allow bootstrap to work it's magic (and the inability to add class extras elsewhere) you kinda have to.

mhvis: This doesn't bother me. The Select2 widgets need separate CSS/HTML and the view is a good place for that. The whole file is concerned with displaying a form using Bootstrap and this fits right in since it's the same issue. I was thinking that we could do the same for the custom Bootstrap select widget

PR #146, opened 2019-05-14
Clean up exception method and expand exception to slot claiming

gwvdh:
I forgot to add the exception to claiming new slots.
Furthermore the exception is now on multiple different kinds of methods, thus move it to a method of User.


mhvis:
Sorry, I had already fixed that in another PR #143 and also changed the methods very similar to what you have :laughing: but I have to wait for @DutcherNL to approve #141 before it'll go into master.

This PR can be merged directly if necessary but it will be changed anyway when #141 is merged

PR #145, opened 2019-05-13
Dining list owners

mhvis:
Todo/done:

- [x] Add `owners` field
- [x] Update `purchaser` field
- [x] Remove `claimed_by` field (including testcases)
- [x] Check for user editing conflicts
- [x] Merge migrations (at the end)
- [x] Do some integration testing

For next PR: add e-mail and telephone number

Edit: in this PR, an association board member always has all rights on all dining lists of the association. For that reason, the owners field is allowed to be blank

Issue #127, opened 2019-05-03
min_diners


mhvis:
Currently, min_diners doesn't do anything other than hiding the delete button (it doesn't block deletion). It's also possible for dining list owners to always change this value afterwards to the highest possible value in order to make deletion possible again, or simply remove diners until deletion is possible again. In that sense it's actually just a constant for the highest possible value and we could just leave it out and replace it by `MAX_SLOT_DINER_MINIMUM`. I'd like to rethink this and possibly remove `min_diners`, because it can be confusing to users because they don't know why they have to provide it and what it is used for, as well as we'd like to have dining list creation and editing as easy as possible, by only presenting useful fields and not confusing someone. (As well as that it can be easily circumvented)

Simple change would be to remove the field and use the `MAX_SLOT_DINER_MINIMUM` constant instead, i.e. always have it set to 6. However I'd propose to rethink this user requirement to something else as there can always be a reason that a dining list is cancelled and thus it should always be possible to cancel regardless of the number of diners.

~I'm thinking of something like this: it should only be possible to delete a dining list if no (manual) sign ups have taken place yet. As soon as someone else has shown interest in the dining list by joining we disable removal and instead change it into a cancel button that marks the dining list as cancelled. Removing all diners would also not re-enable deletion. Haven't thought this through fully yet though.~

Later edit: probably better would be to completely not allow deletion so that dining lists can only be marked as cancelled, that's still easier for users since there are no 2 different concepts of deletion and cancellation, and cases where a dining list must be deleted will be seldom and can easily be handled manually via the admin interface. E.g. when a user made a test dining list and directly cancelled it afterwards we can manually remove it or even just leave it since it wouldn't bother anyone.

DutcherNL:
Due to all the processes involved with dining lists, tracking user help statistics, creating transactions, tracking how many members ate, and the limited amount of daily slots, I'm not really a fan of allowing dining lists to be cancelled. Yes, we can rewrite all of those, but why would we.
Yes, there could be other cases where a large dining list can be cancelled - we can always do that in the admin or something or allows boards to do that on a new page in their association screen (as there are already plans to allow associations to change owners of dining lists) - but the intention of min-diners was never solely the intention of when users can and can not delete dining lists, but to communicate when dining lists could be deleted if too few people are present.
That said, I'm game if we remove min-dining and say in the guidelines that dining lists can be deleted by owners as long as less than 6 members are on the dining list or by board members in special cases.

[...]

PR #150, opened 2019-05-31
Dining list owners

mhvis:
Allow multiple owners of a dining list and check for concurrent editing conflicts.

Currently in this PR, an association board member always has all rights on all dining lists of the association. I'm not sure if that's a good idea, I maybe rather leave that out.

 (Replaces #145)


assets/templates/dining_lists/dining_slot_diners.html
```diff
@@ -82,29 +94,30 @@
...
-                                                {% if entry.has_cleaned %} Clean {% endif %}
+                                                {% if entry.has_cleaned %}Clean{% endif %}
                                             </div>
                                         {% endif %}
                                     {% else %}
+                                        {# External entries have no shop/cook/clean stats #}
                                         <div class="col-9"></div>
                                     {% endif %}
-                                    {% if can_edit_pay %}
+                                    {% if dining_list|is_owner:user %}
```
DutcherNL: If the right to edit payments and cooking stats is the same, there is no need any more for the current seperation and the check can occur only once. Ie. If owner: display 4 clickable buttons, else, display div results.
That said, currently the shop, cook and clean stats can be altered up to 2 days after teh dining list (in order to prevent abuse of help percentage through old dining lists). However, payment is open longer as some people are notoriously slow in paying shit. Maybe we can incorporate that in the layout?

assets/templates/dining_lists/dining_slot_info.html
```diff
@@ -70,15 +85,17 @@ <h3>{{ dining_list.dish }}</h3>
...
             <div class="col-md">
                 {% url 'slot_change' day=date.day month=date.month year=date.year identifier=dining_list.association.slug as url %}
                 <a class="btn btn-primary btn-block" href="{{ url }}">
                     <i class="fas fa-tools"></i> Change information
                 </a>
             </div>
-            {% if can_remove_list %}
+            {% if dining_list|is_owner:user and dining_list.diners.count < dining_list.min_diners %}
+                {# The min_diners check is only performed here and not while a list is actually deleted, so it's #}
+                {# possible to delete a list even if there are more diners than min_diners XD #}
```
DutcherNL: Haha, nice catch.
Speaking of which, I have been toying with the idea of setting it fixed to 6 instead of user-determined as it rarely is an issue and thus specefied by the owner. What do you think?

assets/templates/dining_lists/snippet_diningslot.html
```diff
@@ -16,12 +16,12 @@
...
-            {{ slot.claimed_by }}
-            {% if slot.get_purchaser != slot.claimed_by %} and {{ slot.purchaser }}{% endif %}
-        {% else %}
-            Slot is unclaimed!
-        {% endif %}</div>
+            {% if slot.main_contact %}
+                {{ slot.main_contact }}
+            {% else %}
+                {{ slot.owners.all|join:', ' }}
+            {% endif %}
```
DutcherNL: I have noticed that when we use multiple names in the slot-snippet, a lot of space will be used for the names, which becomes a visual issue (or heavy annoyance) as it pushes away a lot of the remainder of the div. Which is why I have a preference for having one name. What we could also try is only use first names (association and the combination of names would likely create little conflict on who the chefs are. In the full info page we can off course display the full names of all users.

Dining/models.py
```diff
@@ -40,6 +40,13 @@ class DiningList(models.Model):
     The following fields may not be changed after creation: kitchen_cost, min_diners/max_diners!
     """
     date = models.DateField()
+
+    """Todo: the date+association combination determines the URL. This makes it impossible to have multiple dining lists
+    of the same association on the same day. Probably need to change that"""
```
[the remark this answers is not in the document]
mhvis: That's not what's meant here, I agree with not allowing associations to claim more than 1 slot per day. This note is therefore also not important, but I meant that the URL layout shouldn't be enforcing that restriction, it should be enforced by validation checks in the code (only), so that it's possible to make exemptions or change rules later if needed. That will also make the url layout simpler.

Dining/models.py
```diff
@@ -79,14 +77,27 @@ class DiningList(models.Model):
...
+    def is_owner(self, user):
+        """Owner is used to denote that someone has all rights. This method is used to specify if someone has all
+        rights. If we for instance would want to give board members all rights, we could have this method return true if
+        the user is a board member of the same association."""
 
-    def is_authorised_user(self, user):
-        return user == self.claimed_by or user == self.purchaser
+        # Give board members always all rights (is probably not really necessary and can probably be removed)
+        if user.groups.filter(pk=self.association.pk).exists():
+            return True
```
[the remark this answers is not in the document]
mhvis: I fully agree and very much like your reasoning :+1: :smile:

assets/templates/dining_lists/dining_day.html
```diff
@@ -72,10 +73,19 @@ <h4 class="my-0 {% if date.upcoming == date %}font-weight-bold{% endif %}">
             {% include 'dining_lists/snippet_diningslot.html' %}
         {% endfor %}
     </div>
-    {% if can_create_slot %}
-        <a href="{% url 'new_slot' year=date.year month=date.month day=date.day %}"
-           class="col-12 btn btn-block btn-primary">Create and claim a new slot</a>
-    {% endif %}
 
+    {% with cant_create_reason=user|cant_create_dining_list_reason:date %}
```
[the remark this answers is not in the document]
mhvis: Wholeheartedly agree :wink: will change

Dining/models.py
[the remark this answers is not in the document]
DutcherNL: We can indeed simpliy it, but personally I like some readability in the url. In the same way that we have e.g. /associations/Knights/members instead of /a/2/m. That way we can share urls in chats and without clicking you could know where links lead to which is imy always a plus.

assets/templates/dining_lists/dining_slot_diners.html
mhvis: Yes sounds good. Won't do it in this PR though

assets/templates/dining_lists/dining_slot_info.html
mhvis: I have earlier written about this here: #127. I suggest you read that and comment what you think, I'm curious to hear your opinion about that :smiley:. The gist is that I had the same idea kinda.

assets/templates/dining_lists/snippet_diningslot.html
mhvis: For Quadrivium we will have name conflicts as we have for instance 3 people named Marieke who all 3 might join dinner on a Thursday. So we can see if it becomes a problem, however if you want to change it to first names I'll do that, I get that it's still possible to find the full names on the detail page.

mhvis:
The latest commit has the changes according to the comments. You can go over it and if you don't have comments I think you can merge this.

assets/templates/dining_lists/dining_slot_diners.html
DutcherNL: That's fine, maybe I'll do it some day.

assets/templates/dining_lists/snippet_diningslot.html
DutcherNL: The first name bases is just for the dining_slot_snippet when no main contact is set. Yes, there could be some disparsity if a Marieke is one of the (multiple) owners and no main contact is added but then you can get more info on the information page.
That said, we can also make main_contact a property like so:
```
Class DiningList(parents):

_main_contact = models.ForeignKey(settings.AUTH_USER_MODEL,....)

@property
def main_contact(self):
   if self._main_contact == None:
      if self.owners.count() == 1
         return self.owners.first()
   return self._main_contact
```
That case, in case no main contact  is set, but there is only one owner it is automatically assumed main contact until more owners are added.

mhvis: > The first name bases is just for the dining_slot_snippet when no main contact is set.

I get that, that's also in my comment. Question remains: do you want me to change it to first names only? I'll make sure that it displays the full name if there's only 1 owner (but not using the `main_contact` property as that changes the semantic meaning of the main contact)

DutcherNL: Yes, please :)

PR #153, opened 2019-07-29
Transaction ui adjustments

### LINKED ISSUE #148: Transaction overview (user) doesn't have pagination displayed

DutcherNL:
Fixes #148 

Added improvements to transaction management; enabled transactions for users


assets/templates/accounts/snippet_credit_history_association.html
```diff
@@ -29,4 +30,25 @@
...
+    <td class="py-1">
+        {% if transaction.type == 2 %}
+            {% if transaction.source_association == association %}
+                <a class="btn btn-outline-primary"
+                        {% if transaction.target_user or transaction.target_association %}
+                   href="{% url 'association_transaction' association_name=association.slug %}?redirect={{ request.path }}&id={{ transaction.id }}">
+                        {% else %}
+                            href="{% url 'credits:transaction_donate' %}?redirect={{ request.path }}&id={{ transaction.id }}">
+                        {% endif %}
+                <i class="fas fa-pen"></i>
```
mhvis: I suggest adding `<span class="sr-only">Whatever the action is</span>`

assets/templates/accounts/snippet_credit_history_association.html
```diff
@@ -29,4 +30,25 @@
...
+                </a>
+            {% endif %}
+        {% elif transaction.type == 3 %}
+            {% with transaction.id|get_dining_slot as slot %}
+                <a class="btn btn-outline-primary"
+                   href="{{ slot.get_absolute_url }}">
+                    <i class="fas fa-search"></i>
```
mhvis: > I suggest adding `<span class="sr-only">Whatever the action is</span>`

assets/templates/accounts/snippet_memberdetails.html
```diff
@@ -14,6 +14,12 @@
         <td>
             {{ credit }}
         </td>
+        <td class="py-1 px-0">
+            <a class="btn btn-block btn-outline-primary"
+               href="{% url 'association_transaction' association_name=association.slug %}?target_user={{ entry.id }}&redirect={{ request.path }}">
+                <i class="fas fa-plus"></i>
```
mhvis: > I suggest adding `<span class="sr-only">Whatever the action is</span>`

assets/templates/credit_management/snippet_credit_history_user.html
```diff
@@ -29,4 +30,26 @@
...
+    <td class="py-2">
+        {% if transaction.type == 2 %}
+            {% if transaction.source_user == request.user %}
+                <a class="btn btn-outline-primary"
+                {% if transaction.target_user or transaction.target_association %}
+                   href="{% url 'credits:transaction_transfer' %}?redirect={{ request.path }}&id={{ transaction.id }}">
+                {% else %}
+                    href="{% url 'credits:transaction_donate' %}?redirect={{ request.path }}&id={{ transaction.id }}">
+                {% endif %}
+                <i class="fas fa-pen"></i>
```
mhvis: > I suggest adding `<span class="sr-only">Whatever the action is</span>`

assets/templates/credit_management/snippet_credit_history_user.html
```diff
@@ -29,4 +30,26 @@
...
+                </a>
+            {% endif %}
+        {% elif transaction.type == 3 %}
+            {% with transaction.id|get_dining_slot as slot %}
+                <a class="btn btn-outline-primary"
+                   href="{{ slot.get_absolute_url }}">
+                    <i class="fas fa-search"></i>
```
mhvis: > I suggest adding `<span class="sr-only">Whatever the action is</span>`

assets/templates/credit_management/transaction_add.html
```diff
@@ -5,8 +5,29 @@
...
+                {% if not chain %}
+                    <div class="col-md-8 mt-1">
+                        <input type="submit" value="Create" class="btn btn-block btn-primary"/>
+                    </div>
+                {% else %}
+                    <div class="col-md-5 mt-1">
+                        <input type="submit" value="Create" class="btn btn-block btn-primary"/>
+                    </div>
+                    <div class="col-md-3 mt-1">
+                        <input type="submit" value="Create More" name="add_another" class="btn btn-block btn-secondary"/>
```
mhvis: Capitalization of 'Create More' is inconsistent with capitalization in other places, e.g. 'Create and claim new slot'

mhvis: I haven't run this locally, but by looking at this file I don't think it's clear for a user what happens when she clicks on 'Create More'. Shouldn't it be something like 'Create and add another'?

CreditManagement/forms.py
```diff
@@ -54,9 +82,51 @@ class Meta(TransactionForm.Meta):
...
+        self.transaction = transaction_obj
+        self.deleter = deleter
+
+    def clean(self):
+        cleaned_data = super().clean()
+
+        if self.transaction.confirm_moment <= timezone.now():
+            raise ValidationError(_('This transaction can no longer be altered'), code='locked')
+
+        # Validate dining list is still open (except for claimant)
```
mhvis: ??

CreditManagement/forms.py
```diff
@@ -40,6 +58,16 @@ class Meta(TransactionForm.Meta):
...
 
+    def clean(self):
+        cleaned_data = super().clean()
+
+        # Do not allow associations to make evaporating money transactons
+        # (not restircted on database level, but it doesn't make sense to order it)
+        if not cleaned_data.get('target_user') and not cleaned_data.get('target_association'):
+            raise ValidationError("Select a target to transfer the money to.")
+
+        return cleaned_data
```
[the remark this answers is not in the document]
mhvis: Nevermind, I now get that that is done to enable the `UserDonationForm`. It's still better I think though to restrict it generally and make exceptions for e.g. the `UserDonationForm` instead of other way around (but just leave it like this)

assets/templates/credit_management/transaction_delete.html
```diff
@@ -0,0 +1,27 @@
...
+{% block content %}
+    <div>
+        <h3>Detele transaction</h3>
+        <form method="post">
+            {% csrf_token %}
+            <p>
+                Are you sure you want to undo the following transaction?
+            </p>
+            <p class="py-4">{{ transaction_text }}</p>
+            <p class="py-2"></p>
```
mhvis: Add something like `mt-3` to the div below instead of a spacer paragraph.

assets/templates/dining_lists/dining_day.html
```diff
@@ -72,10 +72,32 @@ <h4 class="my-0 {% if date.upcoming == date %}font-weight-bold{% endif %}">
...
+        <div class="col-12 d-md-none d-inline">
+            {% if can_create_slot %}
+                <a href="{% url 'new_slot' year=date.year month=date.month day=date.day %}"
+                   class="btn btn-primary col-12">Create and claim a new slot</a>
+            {% endif %}
+            <a class="btn btn-secondary col-5 mt-2"
+               href="{% url 'credits:transaction_donate' %}?redirect={{ request.path }}">
+                <i class="fas fa-coins"></i> Single use
+            </a>
+        </div>
```
mhvis: Here you have `col`s inside of `col`s while `col`s should be inside `row`s

assets/templates/dining_lists/dining_day.html
```diff
@@ -72,10 +72,32 @@ <h4 class="my-0 {% if date.upcoming == date %}font-weight-bold{% endif %}">
...
-    {% if can_create_slot %}
-        <a href="{% url 'new_slot' year=date.year month=date.month day=date.day %}"
-           class="col-12 btn btn-block btn-primary">Create and claim a new slot</a>
-    {% endif %}
+
+    <div class="row">
+        <div class="btn-group col-12 d-md-inline-flex d-none">
+            <a class="btn btn-secondary col-3"
+               href="{% url 'credits:transaction_donate' %}?redirect={{ request.path }}">
+                <i class="fas fa-coins"></i> Single use
```
mhvis: I don't think it's clear for a user what this is for. Maybe use something like 'Kitchen use without creating dining list'. Or you could add some explanation below in small font.

mhvis: The gist is: the user should not need to go to the help page to have UI actions explained

assets/templates/dining_lists/dining_day.html
mhvis: Some suggestions:
* col-12/row cancel out
* content should be inside cols, e.g. `btn-group` should be on a div inside the col div, not on the same div

mhvis:
Nice work in general!

CreditManagement/models.py
```diff
@@ -183,6 +185,15 @@ def source(self):
...
         return self.target_association if self.target_association else self.target_user
 
+    def get_class_type(self):
+        if self.type == FixedTransaction.type_code:
+            return FixedTransaction
+        if self.type == PendingTransaction.type_code:
+            return PendingTransaction
+        if self.type == PendingDiningTransaction.type_code:
+            return PendingDiningTransaction
+        raise KeyError("Stored key does not match set keys")
```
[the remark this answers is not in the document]
DutcherNL: Nope, in the queryset that combines all transaction types into 1 queryset I use Union(). Union however has the side effect that python interprets all objects in the queryset of the first type mentioned in the union. Thus union(Fixed, Pending, PendingDining) results in one queryset interpreted of FixedObjects.
Alas

CreditManagement/views.py
```diff
@@ -15,44 +16,201 @@ class TransactionListView(ListView):
...
 
     def get_queryset(self):
-        return AbstractTransaction.get_all_transactions(user=self.request.user).order_by('-pk')
+        return AbstractTransaction.get_all_transactions(user=self.request.user).order_by('-order_moment')
 
 
-class TransactionAddView(LoginRequiredMixin, View):
-    template_name = "credit_management/transaction_add.html"
-    context = {}
+class CustomAccessMixin:
```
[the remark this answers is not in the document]
mhvis: The message given with HttpResponseForbidden is not displayed to the user.

CreditManagement/views.py
```diff
@@ -15,44 +16,201 @@ class TransactionListView(ListView):
...
+    def check_access_permission(self, request):
+        transaction_id = request.GET.get('id', None)
+
+        if transaction_id is not None:
+            # get the instance
+            t_order = get_object_or_404(PendingTransaction, pk=transaction_id)
+            if t_order.source_user != request.user:
+                return HttpResponseForbidden("You do not have access to this transaction")
+            if t_order.confirm_moment <= timezone.now():
+                return HttpResponseForbidden("This transaction can no longer be altered")
```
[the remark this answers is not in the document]
mhvis: > The HTTPResponseForbidden yields a 403 page where the error is presented.

You might think that but I'm pretty sure that that's not the case.. You see the message only because you have DEBUG=True in your settings. On the production server though the DEBUG setting is False which will ensure that only the literal text '403 Forbidden' will be shown to the user (this is mentioned in the Django docs..). It's easy to create a custom 403/404 page, by adding 403.html/404.html to the root template folder. However I don't want that as it's totally unnecessary. A user should never end up on a 403 or 404 page via UI interactions, so only via other means and I don't want views that are devoted to guiding people that are not using the 'official UI'

CreditManagement/views.py
```diff
@@ -15,44 +16,201 @@ class TransactionListView(ListView):
...
-            form = AssociationTransactionForm(association, request.POST)
+            if not self.request.user.is_board_of(association.id):
+                return HttpResponseForbidden("You do not have access to this transaction")
+            t_order = get_object_or_404(PendingTransaction, pk=transaction_id)
+            if t_order.confirm_moment <= timezone.now():
+                return HttpResponseForbidden("This transaction can no longer be altered")
         else:
-            form = UserTransactionForm(request.user, request.POST)
+            if not self.request.user.is_board_of(association.id):
+                return HttpResponseForbidden("You can not make transactions for this association")
```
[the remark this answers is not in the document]
mhvis: > But you know, when someone starts looking for leaks...
???

This kind of insinuates that you think that I suggested that HttpResponseForbidden is not really necessary just because it's not possible to get there by following links. Do you really think that that's what I meant?...

Always return HttpResponseForbidden, but the message provided is only for programming debugging reasons and (as I explained above) is not displayed to the user, nor should it be. We should not be wasting code on guiding people who are not using the UI.

CreditManagement/views.py
```diff
@@ -22,62 +22,71 @@ def get_queryset(self):
...
 
     def post(self, request, *args, **kwargs):
-        transaction_obj = self.get_transaction()
-        form = self.get_form(transaction=transaction_obj, data=request.POST)
+        form = self.get_form(transaction=self.transaction_obj, data=request.POST)
 
         if form.is_valid():
             form.save()
-            if transaction_obj is None:
+            if self.transaction_obj is None:
```
mhvis: Using different views for creation and updating of transactions makes the code much cleaner (no need for `if transaction is None` in a lot of places). See e.g. the Django built in generic editing views.

CreditManagement/views.py
```diff
@@ -96,20 +105,28 @@ def post(self, request, *args, **kwargs):
...
+            if self.transaction_obj.confirm_moment <= timezone.now():
+                return HttpResponseForbidden("This transaction can no longer be altered")
+        return super(TransactionBaseAlterView, self).check_access_permission(request)
+
 
-class TransactionUserView(LoginRequiredMixin, TransactionAlterView):
-    def get_form(self, transaction=None, data=None):
+class TransactionUserView(LoginRequiredMixin, TransactionBaseAlterView):
+    """
+    Basic user transaction view
```
mhvis: I've learned to be using a standard consistent coding style as well as a consistent docstring style at my summer job where we're also using Django, and it makes the code much nicer and homogeneous. I want to apply this here as well (#157)

For docstring style I want to be following the Google style guide ([here](https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings)). I've compared the different styles that there are out there and this is a clear winner over the default style used by PyCharm. You can change the docstring format to Google's in PyCharm settings.

I would suggest that if you add comments that you should make them meaningful and not just repeat the name of the class.

CreditManagement/views.py
mhvis: Come to think of it, there's no need whatsoever to allow transactions to be altered. It can lead to really confusing situations for users.... E.g. someone on the receiving end would expect that his balance is fixed and cannot go down without interaction by himself. With this implementation that's however not the case (if I understand correctly) as a transaction sender might later alter a transaction.

It seems often that you think that more functionality is better for the user. It's really not. You can find a lot online about what constitutes good UX, e.g. https://www.atlanticbt.com/insights/functionality_vs_usability_ux/. I often need to argue a lot about this while for me it's usually pretty clear, and it's getting annoying.

assets/templates/dining_lists/dining_day.html
```diff
@@ -75,44 +75,39 @@ <h4 class="my-0 {% if date.upcoming == date %}font-weight-bold{% endif %}">
...
+                    <div class="btn-group col-12 d-md-inline-flex d-none">
+                        <a class="btn btn-secondary col-3"
+                           href="{% url 'credits:transaction_donate' %}?redirect={{ request.path }}">
+                            <i class="fas fa-coins"></i> Single use
+                        </a>
+                        {% if can_create_list %}
+                            <a href="{% url 'new_slot' year=date.year month=date.month day=date.day %}"
+                               class="btn btn-primary">Create and claim a new slot</a>
+                        {% elif cant_create_reason %}
+                            <div class="btn btn-primary disabled">{{ cant_create_reason }}</div>
```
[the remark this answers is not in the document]
mhvis: Sorry but I really don't get that when you change such a thing why you don't check for other instances that need to be changed as well, e.g. on line 89. It makes the review process really tedious for me when I have to go over each thing so meticulously..

I kind of want to stop doing this so I'll probably go think about other ways to be involved. One way could be that you let me handle the implementation..

assets/templates/credit_management/transaction_add.html
```diff
@@ -23,7 +23,7 @@ <h3>Create new transaction</h3>
                         <input type="submit" value="Create" class="btn btn-block btn-primary"/>
                     </div>
                     <div class="col-md-3 mt-1">
-                        <input type="submit" value="Create More" name="add_another" class="btn btn-block btn-secondary"/>
+                        <input type="submit" value="Create and add more" name="add_another" class="btn btn-block btn-secondary"/>
```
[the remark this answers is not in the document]
mhvis: 'Create and add more' can mean multiple things, e.g. it's not indicated how many more, 'more' is a broad term... 'Create and add another' is way more specific. It's important for good UX to be as specific as possible. It's annoying for users if it's not exactly clear what's going to happen. It most surely outweighs the extra 4 characters that are needed.

assets/templates/dining_lists/dining_day.html
```diff
@@ -75,44 +75,39 @@ <h4 class="my-0 {% if date.upcoming == date %}font-weight-bold{% endif %}">
...
-                   class="btn btn-primary">Create and claim a new slot</a>
-            {% endif %}
-        </div>
+        {% if date|dining_list_creation_open %}
+            {% with user|can_create_dining_list:date as can_create_list %}
+                {% with user|cant_create_dining_list_reason:date as cant_create_reason %}
+                    <div class="btn-group col-12 d-md-inline-flex d-none">
+                        <a class="btn btn-secondary col-3"
+                           href="{% url 'credits:transaction_donate' %}?redirect={{ request.path }}">
+                            <i class="fas fa-coins"></i> Single use
```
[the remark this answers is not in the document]
mhvis: 'Donate to kitchen' sounds like the kitchen is a charity that you can give money to. I would say something like 'Use of kitchen without a dining list'. Not exactly that but in that direction. The main button action label should be short and most important concise and to the point. Now it's likely not possible to make it fully clear in the short button label so in that case you should add some small text below the button label explaining the action in more detail.

CreditManagement/views.py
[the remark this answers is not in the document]
DutcherNL: I just checked with both debug on and off. I do get the message. Just a blank page with the text "You do not have permission"/"Transaction expired" etc...

CreditManagement/views.py
DutcherNL: I'll try to use it in my future works.

CreditManagement/views.py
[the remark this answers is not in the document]
mhvis: Okay, still think it's better to not be helping/guiding people who are not using the UI.

assets/templates/accounts/snippet_credit_history_association.html
```diff
@@ -39,14 +39,14 @@
                         {% else %}
                             href="{% url 'credits:transaction_donate' %}?redirect={{ request.path }}&id={{ transaction.id }}">
                         {% endif %}
-                <i class="fas fa-pen"></i>
+                <span class="fas fa-pen"></span>
```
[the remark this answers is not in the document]
mhvis: > You can use the Bootstrap sr-only class for that.

PR #159, opened 2019-08-08
Mailed messages

DutcherNL:
Implemented full urls
Small mail layout improvements

### DIFF
--- a/Dining/views.py
+++ b/Dining/views.py
@@ -306,15 +306,17 @@ class EntryDeleteView(LoginRequiredMixin, SingleObjectMixin, View):
             template = "dining/dining_list_entry_removed_by"
             context = {'entry': entry, 'dining_list': entry.dining_list, 'remover': request.user}
 
-            # Send mail to the people on the dining list
-            send_templated_mail(subject=subject, template_name=template, context_data=context, recipient=entry.user)
-
             success_msg = "The user is removed from the dining list"
 
         # Process deletion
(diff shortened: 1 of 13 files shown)


Dining/views.py
```diff
@@ -306,15 +306,17 @@ def post(self, request, *args,**kwargs):
...
-            send_templated_mail(subject=subject, template_name=template, context_data=context, recipient=entry.user)
-
             success_msg = "The user is removed from the dining list"
 
         # Process deletion
         form = DiningEntryDeleteForm(entry, request.user, {})
         if form.is_valid():
             form.execute()
+
+            if entry != request.user:
```
mhvis: I think this is wrong

Dining/views.py
```diff
@@ -306,15 +306,17 @@ def post(self, request, *args,**kwargs):
...
+                # Send mail to the people on the dining list
```
mhvis: It's always only 1 person right? (Namely the entry that is deleted)

Dining/views.py
```diff
@@ -587,7 +593,7 @@ def delete(self, request, *args, **kwargs):
             messages.success(request, _("Dining list is deleted"))
 
             # Need to use reverse from the DiningListMixin superclass
```
mhvis: Remove this comment

Dining/views.py
DutcherNL: Yup. Oops. That was too quick a fix haha

PR #163, opened 2019-09-16
Allow ownership removal of self

mhvis:
Very minor change to allow people to remove themselves as the owner

### DIFF
--- a/Dining/views.py
+++ b/Dining/views.py
@@ -520,9 +520,6 @@ class SlotInfoChangeView(LoginRequiredMixin, SlotMixin, TemplateView):
             payment_form.save()
             messages.success(request, "Changes successfully saved")
 
-            # Ensure that current user remains owner of the dining list
-            self.dining_list.owners.add(request.user)
-
             return HttpResponseRedirect(self.reverse('slot_details'))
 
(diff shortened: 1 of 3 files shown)


Dining/views.py
```diff
@@ -478,51 +478,26 @@ def post(self, request, *args, **kwargs):
...
+        # Dispatch
         return super().dispatch(request, *args, **kwargs)
 
 
-class SlotOwnerMixin(LoginRequiredMixin, SlotMixin, IsOwnerMixin):
-    """
-    Combined mixin for slot pages that should only be accessable for dining list owners
-    """
-    pass
-
```
DutcherNL: The reason I implemented it like this is otherwise, when calling the SlotownerMixin the LoginRequiredMixin must also be called in every child in order to prevent an error in running the code to confirm the logged in user is owner. Arguably this is also the case for SlotMixin as that creates visittrackers and could be moved there so could be better there.

Anyhow, the current removal does not address the lack of loginchecks, either revert the LoginRequired back to each subchild or implement it in some way in the parent. As the order of parent processing is important I propose something similar to what I did above, though I'd love to hear your opinion on that.

PR #164, opened 2019-09-18
Pipenv and linting tools

### LINKED ISSUE #157: Coding style
While not directly important or high priority, it does make me happy to have the code follow a coding style and have this enforced. This holds as well for the docstrings, so that we even might be able to use a documentation generator in the future. While I haven't investigated it yet, it would likely become the Google style guide and add tools like yapf, pylint and pydocstyle. However I'll only implement these when you're happy with that and also probably not very soon.

mhvis:
This makes the switch from `requirements.txt` to using Pipenv for the dependencies. I've been using it for other projects now and can recommend it, it's really nice. But you can try for yourself and we don't need to switch if you don't want.

I removed the dependency version requirement, since Pipenv uses dependency locking that's not really needed anymore. So we'll automatically use the latest dependency versions.

Next to that this adds linting tools to the development dependencies and even runs the linter in the build process, so we'll have to fix all linting errors now before the build will succeed again.

Fixes #157


mhvis:
Closed as build fails

PR #171, opened 2019-10-12
Make max_diners optional

### LINKED ISSUE #147: Max diners
Suggestion: make max diners optional so that people don't have to put down a large number if there is no max

mhvis:
Branched off from #170, first merge that one before looking into this one.

Fixes #147


mhvis:
Don't think that this is very useful, would like to cancel this PR and close the corresponding issue

PR #197, opened 2021-10-25
Major overhaul

mhvis:
All changes, in no particular order:

* Diners contact info: e-mail and telephone number. A user can enter their contact info if they agree with the privacy implications, if they do so the contact info will be shown to everyone who subscribed to the same dining lists. It is shown on the diners page.
* Redesign of the following pages:
  * Dining info
  * Dining overview
  * Diners list (stats as buttons instead of checkboxes which save directly)
* Dining list deletion: it is only possible to delete when no diners are signed up. This prevents that the site has to reimburse the kitchen cost of all diners and notify them, instead the cook has to do this explicitly first by removing all diners. The edit page includes a small description on how to cancel or delete a dining list.
* Grocery cost splits: a more explicit way to split grocery costs over diners. When a cook splits the costs, all diners will be notified with the payment link and contact info. Diners can optionally choose to pay using their account balance, if the cook has enabled this option.
* Removed the negative balance exception for Quadrivium (!). This makes checking for sufficient balance more transparent, we no longer need to deal with and allow (potentially large) negative balances for Quadrivium members. The new approach is that Quadrivium members can access a page where they can upgrade their balance by specifying an amount and checking a box to confirm. The Quadrivium treasurer has access to a page which shows the total upgraded amount for each user and can debit those users.
* Removed the "news and updates" page because the content is outdated and more text means less incentive for users to actually read it. Instead if we need to inform the users of new features or clarify unclear features we should just add a notice at the appropriate places (near the actual feature) and later remove the notice when no longer necessary.
* Update Bootstrap, Font Awesome, Python libraries.
* Bugfixes (see linked issues)

To do for merging:

* Need to inform the Quadrivium treasurer on the new invoicing approach. (Maarten)
* Need to check if board agrees with the ability to pay using account balance. (This ability can be easily turned off)
* The PR became a bit too huge, some of these changes may not be desired. I can undo unwanted changes in this PR or split them to a separate PR.

### DIFF
--- a/assets/templates/dining_lists/day.html
+++ b/assets/templates/dining_lists/day.html
@@ -3,6 +3,8 @@
 {% load humanize %}
 {% load dining_tags %}
 
+{% block title %}{{ block.super }} {{ date|date:"SHORT_DATE_FORMAT" }}{% endblock %}
+
 {% block content %}
     <div class="row mb-3">
         {# Button toolbar #}
@@ -12,7 +14,7 @@
                     {# (upcoming is a class method, i.e. not specific to the instance) #}
                     {% with u=date.upcoming %}
                         <a href="{% url 'day_view' day=u.day month=u.month year=u.year %}"
-                           class="btn btn-block btn-outline-primary">
+                           class="btn btn-block btn-outline-primary {% if u == date %}disabled{% endif %}">
                             {% if u == date.today %}Today{% else %}Upcoming{% endif %}
                         </a>
                     {% endwith %}
--- a/assets/templates/dining_lists/dining_add.html
+++ b/assets/templates/dining_lists/dining_add.html
@@ -1,13 +1,27 @@
 {% extends 'base.html' %}
 
+{% block title %}{{ block.super }} {{ date|date:"SHORT_DATE_FORMAT" }}{% endblock %}
+
 {% block content %}
-    <div class="text-center">
-        <h4>{{ date|date:DOW_DATE_FORMAT }}</h4>
-    </div>
-    <h3>Create new dining list</h3>
+    <p>
+        <a href="{% url 'day_view' day=date.day month=date.month year=date.year %}">
+            ← Back to day view
+        </a>
+    </p>
+    {#    <div class="text-center">#}
+    {#        <h4>{{ date|date:DOW_DATE_FORMAT }}</h4>#}
+    {#    </div>#}
+    <h3>New dining list</h3>
     <form method="post" action="{% url 'new_slot' year=date.year month=date.month day=date.day %}">
         {% csrf_token %}
+        <div class="form-group row">
+            <label for="date" class="col-sm-2 col-form-label">Date</label>
+            <div class="col-sm-10">
+                <input type="text" class="form-control-plaintext" id="date"
+                       value="{{ date|date:"SHORT_DATE_FORMAT" }}">
+            </div>
+        </div>
         {% include "snippets/bootstrap_form.html" with form=slot_form horizontal=True %}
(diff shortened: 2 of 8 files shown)


LenaWil:
Is this one still alive?

LenaWil:
> Removed the negative balance exception for Quadrivium (!). This makes checking for sufficient balance more transparent, we no longer need to deal with and allow (potentially large) negative balances for Quadrivium members. The new approach is that Quadrivium members can access a page where they can upgrade their balance by specifying an amount and checking a box to confirm. The Quadrivium treasurer has access to a page which shows the total upgraded amount for each user and can debit those users.
I am not the Q treasurer, but the bill is often sent at the end of the year, right?
How exactly would this make a difference? Instead of negative balances, you now just get negative balances, but visible in a different place. It also just seems, as a Q member, but not a treasurer, just more hassle to upgrade the balance every time, where it’s now just done automatically.

mhvis:
> Is this one still alive?

It became too big to merge and I'm not really active anymore with the project. But I'll try to split this into smaller PRs.


> > Removed the negative balance exception for Quadrivium (!). This makes checking for sufficient balance more transparent, we no longer need to deal with and allow (potentially large) negative balances for Quadrivium members. The new approach is that Quadrivium members can access a page where they can upgrade their balance by specifying an amount and checking a box to confirm. The Quadrivium treasurer has access to a page which shows the total upgraded amount for each user and can debit those users.
> 
> I am not the Q treasurer, but the bill is often sent at the end of the year, right? How exactly would this make a difference? Instead of negative balances, you now just get negative balances, but visible in a different place. It also just seems, as a Q member, but not a treasurer, just more hassle to upgrade the balance every time, where it’s now just done automatically.
> 
> Edit: although I do have to admit, I haven’t read the code yet, maybe I have the wrong idea on how it works.

It's a semantic difference. Currently each time money is withdrawn a check has to be done if that person is a member of Q, to check if the balance can be negative. It is nice to get rid of that check, for instance using a more explicit upgrade form. Then code that deals with money does not have to take negative balances into account. But if you have a better idea than a form, let me know. Having it more explicit also makes it easier for members and the treasurer to understand. And if there would be multiple associations that uses invoicing, it allows the person to choose the association if they are member of both.

LenaWil:
> for instance using a more explicit upgrade form.
Again, haven’t tested the code so far, but it would be great if there is an option to just select Q or other associations that allow that in the settings form and that form be submitted in automatically for the members for members of Q, so I don’t have to up my balance every time I want to subscribe to a dining list, and/or don’t have money in there that doesn’t need to be there. 
(If I leave Scala and there is still money in there, that has been on my Q-bill, I would have to get it back somehow.)

LenaWil:
> But I'll try to split this into smaller PRs.
Thanks!

mhvis:
> Again, haven’t tested the code so far, but it would be great if there is an option to just select Q or other associations that allow that in the settings form and that form be submitted in automatically for the members for members of Q, so I don’t have to up my balance every time I want to subscribe to a dining list, and/or don’t have money in there that doesn’t need to be there. (If I leave Scala and there is still money in there, that has been on my Q-bill, I would have to get it back somehow.)

An automatic balance increase for Q members at sign up time would be a way to make it easier. The goal is to make the Q balances behave the same way as the other associations in the background, and directly move the negative balance to a separate boekhoudaccount for Quadrivium. But that might indeed be possible to do automatically without an explicit form, so that the balance for Q members just remains 0.

However as I'm less active I probably won't write that. And most of this PR including this Q change will probably not make it to the master branch.

PR #221, opened 2022-12-21
Update (dev-)requirements.txt

LenaWil:
I had some problems getting this to run, and they went away when I ran pip-compile on both files and then installed the requirements again. 

Whether or not those are related, I don’t know, but it can’t hurt to run it. 

I didn’t use the upgrade flag.


mhvis:
Thanks!

@DutcherNL I'll leave merging to you.

PR #225, opened 2022-12-22
Fix duplicate requirements file

mhvis:
Remove the duplicate requirements in the `requirements` folder and fix the GitHub codecov action.

I'm all for moving the requirements into a separate folder, but this was the easiest fix for now.


.github/workflows/codecov.yml
```diff
@@ -10,7 +10,7 @@ jobs:
       DJANGO_ENV: 'TESTING'
     strategy:
       matrix:
-        python: ["3.7"]
+        python: ["3.10"]
```
LenaWil, line 13: Is there a reason to not use python 3.11 here?

mhvis: No particular reason, I'll change it to 3.11 (as long as Django supports it which is the case).

It should be the same version as used on the server. The Docker image in `Dockerfile` is used on the server. So I'll update it there as well.

mhvis: Actually that Python version is changed in #224. I'll wait till that one is merged @DutcherNL :smile:

DutcherNL (approved):

PR #226, opened 2022-12-22
Fix tests and linter comments

mhvis:
The tests in `utils/tests/test_form_validation.py` did not execute because they were not found by the test discovery. The test discovery only looks in Python ~~modules~~ modules and packages, and because the folder `utils` did not have `__init__.py` it was not a valid ~~module~~package.

Some tests that did not execute before now failed. After fixing, other tests failed that depended on it. These have been fixed as well.

Most changes in this PR are for formatting and the rest are the fixed test cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,3 +32,12 @@ inside the virtual environment
 and run `pip-compile requirements.in`.
 See [pip-tools documentation](https://github.com/jazzband/pip-tools)
 for details.
+
+## Code style
+
+The linter Flake8 expects the code, including docstrings, to follow the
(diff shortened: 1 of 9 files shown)


DutcherNL:
Thanks for the fix. Though I disagree with the renaming convention of the assertFormValid methods. Not that I want methods with Camelcase all over, but because of continuity with the python unittest module which Django expands in which assertX uses CamelCase naming convention, thus for continuity assert methods used this convention as well.

That said, this PR is important enough not to delay

PR #228, opened 2022-12-27
Fix payment mail button

### LINKED ISSUE #166: Send mail button disables before post request is send on some mobile browsers
It works fine for desktops, though nothing happens in some Chrome browsers on Android mobiles.

mhvis:
Fixes #166.


dining/views.py
```diff
@@ -600,6 +605,31 @@ def form_invalid(self, form):
         return HttpResponseRedirect(self.get_success_url())
 
     def form_valid(self, form):
-        form.send_reminder(self.request)
-        messages.success(self.request, "Diners have been informed")
-        return super(SlotPaymentView, self).form_valid(form)
+        # We use a critical section to prevent multiple emails from being sent.
```
DutcherNL, line 608: Arguably the logic below should be in the form as well instead of the view. Retrieve the error thrown by the form and communicate the message displayed based on the error-code the form has. This would be more in line with the division that Django made for view and form logic. That way we can also 'call' the form elsewhere if at any point things are moved around.

Also, it would be nice if you could write a test-case (I've really learned the importance of testcases over the lockdown) to catch the required behavior.

DutcherNL (changes requested):

mhvis:
I have added unit tests for `SendReminderForm`. I noticed later that you already had tests for it. I prefer to keep mine as well because they are shorter and more self-contained.

Tips for future tests:

* Try not to use fixtures and test mixins. Not relying on fixtures or mixins make the tests more self-contained, and is more readable and maintainable. (This is loosely quoted from the Django docs.)
* Writing tests is also a good way to restructure the code that is being tested such that each method does 1 thing which can be tested directly. In this case, by splitting off the retrieval of entries in a separate method we can just test that the correct entries are returned. Then we don't need to mock which is way more readable and shorter. In this case, mock also makes the test case implementation dependent. The test case fails when we change the send implementation, which is the case here.

The rate limiting implementation depends on PostgreSQL. The test is disabled when the test runner uses SQLite. We need to change the GitHub Action so it uses PostgreSQL instead of SQLite.

If you accept the changes, can you let me do the merge? Then I can practice squashing the commits and doing a fast-forward merge instead of having a merge commit created.

dining/tests/test_forms_sendreminderform.py
```diff
@@ -0,0 +1,157 @@
...
+
+class SendReminderFormLockTestCase(TransactionTestCase):
+    """Tests that no simultaneous emails are sent.
+
+    This needs to inherit from TransactionTestCase because we are testing
+    code with a database transaction.
+    """
+
+    @skipUnlessDBFeature('has_select_for_update', 'has_select_for_update_nowait')
+    def test_send_reminder(self):
```
mhvis, line 118: Dat codecov dit niet herkend komt doordat de test runner SQLite gebruikt.

DutcherNL (approved):

PR #231, opened 2022-12-27
Contact info

### LINKED ISSUE #138: Telephone numbers
Feature request from someone else: allow telephone numbers in the system so that people can be stalked about dinner and payment.

If we make it optional and note down that everyone can see it if you fill it in it should be reasonably fine for privacy reasons I think.

mhvis:
Adds fields for contact info (phone number and email) to the user profile settings. The contact info is visible to everyone who subscribed to the same dining list.

Closes #138.

WIP: I have to wait for the other PRs to be merged/rejected.

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -5,3 +5,5 @@ django-autocomplete-light
 environs[django]
 whitenoise  # Used for static files in production
 Pillow  # This one can possibly be removed
+
+django-phonenumber-field[phonenumbers]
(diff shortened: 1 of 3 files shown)


mhvis:
I might revive this PR if people think it's a good idea.

The idea: people who are subscribed to a dining list are able to view the e-mail address and optionally phone number (if provided) of the other diners. This kind of helps to fix #290. I don't see privacy issues here, when there is a clear notice next to the field.

PR #234, opened 2023-01-02
Workflow simplications and use Postgres database for CI

mhvis:


mhvis:
I kind of want to turn off the codecov comment, but keep the rest from codecov. Then all coverage info can still be found under the checks list. If you want to keep the comment I'll update the PR.

DutcherNL (approved):

PR #247, opened 2023-03-08
Case-insensitive validation of username uniqueness

mhvis:
This is a bit tricky.

Django by default considers 'asdf' and 'Asdf' as different users. The Allauth library that we use considers them as the same user. This leads to that when two users exist with the same username but different case, they cannot login. Instead they get a server error.

This PR is a quick fix. After merging we should manually fix the duplicate usernames that already exist (there are 2 I think).


DutcherNL (approved):

mhvis:
There were actually already 7 duplicate users in the database, who were not able to log in. I renamed the duplicate usernames.

PR #250, opened 2023-03-20
Fix dining list can not be adjusted

mhvis:
This PR removes the check for `is_adjustable` from the DiningList clean method.

I have checked and it is safe to do so. Each form has their own check on `is_adjustable` when it is applicable. It should be on forms anyway instead of on the model clean method, because some forms should be able to modify the dining list after the adjustable period.

### DIFF
--- a/dining/models.py
+++ b/dining/models.py
@@ -82,9 +82,6 @@ class DiningList(models.Model):
     is_adjustable.boolean = True
 
     def clean(self):
-        # Validate dining list can be changed.
-        if self.pk and not self.is_adjustable():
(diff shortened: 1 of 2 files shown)


mhvis:
Addition:

This PR is a response to a bug report from someone. They indicated that it's not possible to change the payment link when the dining list is more than 2 days old. That is a bug. This PR also enables changing the dish, max diners and sign up deadline after the adjustable time. That's not a problem though. It's not possible to reopen an old dining list by setting the sign up deadline in the future, because the sign up deadline can't be later than the date of the dining list.

DutcherNL (approved):

PR #254, opened 2023-05-01
Increase maximum length of payment link

mhvis:
This fixes an issue that was reported by mail. The maximum length of the payment link is too small. Some payment request links require more than 100 characters.


dining/models.py
```diff
@@ -56,7 +56,8 @@ class DiningList(models.Model):
 
     auto_pay = models.BooleanField(default=False)
 
-    payment_link = models.CharField(blank=True, max_length=100, help_text="Link for payment, e.g. a Tikkie link.")
+    # Why max_length=2000? -> https://stackoverflow.com/q/417142/2373688
```
LenaWil, line 59: Do we still support IE/pre-chromium edge?

Not that I mind the limit and we need to pick some limit, just wondering, since they are the only browser not supporting it.

mhvis: No you don't have to care about IE. We just need to pick some limit indeed. But I do think that the banks will honor this IE limit and not create payment links that are longer than 2000 characters, so that these payment links still work in IE. But I'm fine with any limit, as long as it is more than 2000.

PR #255, opened 2023-05-01
Layout change on dining info page, and truncate the payment URL

### LINKED ISSUE #243: long payment link causes layout issues on Firefox mobile
Screenshot_20230227-193545

mhvis:
This is a layout change on the dining info page. Feedback is welcome. I don't mind changing it further or keeping the old layout.

The primary reason for the change is to have the info below each other instead of to the right, on mobile. That works better on mobile.

I'm not sure about the icons. Maybe it's nicer without icons.

Fixes #243 

Before:
image

After:
image

### DIFF
--- a/assets/templates/dining_lists/dining_slot_info.html
+++ b/assets/templates/dining_lists/dining_slot_info.html
@@ -1,5 +1,5 @@
 {% extends 'dining_lists/dining_slot.html' %}
-{% load dining_tags %}
+{% load dining_tags credit_tags humanize %}
 
 {% block tab_info %}active{% endblock %}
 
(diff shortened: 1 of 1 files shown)


LenaWil:
I don’t know enough about Bootstrap and the CSS classes that are used in this. I could blindly test and approve it, but I am not sure how good the quality of the review would be?

mhvis:
Quality of HTML/CSS does not matter (HTML is not code). Only thing that matters is whether it looks nice and has good UX. You can use any CSS classes you like when you're writing something. You can also go wild and write your own CSS and not use Bootstrap, I don't see any issue in that either.

DutcherNL (approved):

PR #257, opened 2023-05-18
Format everything with black

LenaWil:
This is what the pull request would look like.


LenaWil:
I do really encounter extremely obscure bugs everytime: https://github.com/pre-commit/pre-commit/pull/2890

mhvis:
I like the consistency very much. The same exact formatting everywhere makes that we only have to focus on functionality instead of formatting. But if @DutcherNL does not like this code style we should not implement this.

I personally am against the use of pre-commit. By running the tools manually it remains transparent what is happening, you learn all capabilities and purposes of the tools and you understand what you are actually committing. That is also why I use `git add` to explicitly add each changed file instead of just committing everything or using a GUI. The end goal should be to eventually write code in such a way that it is already fully linted. Then the tools are no longer necessary and the GitHub Action will ensure that you did not forget anything.

Having said that, the use of pre-commit is optional so their config files can stay.

LenaWil:
Yeah, okay, it’s a proof of concept, and I have to say I’m less inclined to use pre-commit at all since discovering that bug. 

But we will see what @DutcherNL thinks about it.

DutcherNL (approved):

PR #263, opened 2023-06-12
Add extra check for pip-compile

LenaWil:
And also accidentally format the YML file the wrong way… again.

Anyway it’s a draft, to look into later.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,29 +26,32 @@ jobs:
           --health-retries 5
 
     steps:
-    - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


mhvis:
Do you want to check if dependencies are up to date?

mhvis:
No don't worry I was just curious and I like discussing random ideas.

LenaWil:
@mhvis Which Python version is running on your server? 3.8?

mhvis:
3.11. The server runs the Docker image which is compiled from the Dockerfile in the root path of this repository. You can also change the version in that file.

PR #264, opened 2023-06-21
Statistics page

mhvis:
There was a request by email to pull some statistics from the database. Now the easiest approach is to add a page for that. So this is a quickly made page to compute dining list and diner counts over a period.

It may be a good idea to restrict the page access, but for now I think having no direct link to the page is sufficient.


mhvis:
I might merge this quickly unless there are remarks.

README.md
```diff
@@ -12,16 +12,12 @@ Cooking website.
...
-- Install Docker and Docker Compose
-- `docker-compose build`
-- `docker-compose up`
-- `docker-compose exec app python manage.py migrate`
+The app can also be built and run with Docker.
 
 ## Development commands
 
 * Lint code: `flake8`
+* Blacken the code `black .`
```
LenaWil, line 20: Could you add isort too?
```suggestion
* Blacken the code `black .`
* Sort the imports with `isort --ac .`
```

LenaWil (commented):
Okay, I am an sich willing to approve this if you want to, but I am in my exam weeks and the code isn’t actually check out on my current device so I can’t actually test the page right now.

README.md
mhvis: Thanks for the suggestion, will do

docker-compose.yml
```diff
@@ -1,33 +0,0 @@
-version: '3.7'
-
-# Docker configuration for development
-#
-# If you need local adjustments, use a Docker compose override file.
```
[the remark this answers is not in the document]
mhvis: I changed my local setup to use Docker for Postgres only, and run the app without Docker. So then I realized that every local setup is different and instead of overriding this compose file it's probably just easier to let someone create their own compose file if they want.

So I still have a compose file locally, with just Postgres, but I'll leave it out of the repo.

It's also possible to run it locally with Docker without a compose file.

The server actually also uses a compose file but that one looks very different.

mhvis:
Thanks for the suggestions. I'll merge it asap so that I can reply on the request. You don't have to approve, we can merge without review.

Issue #265, opened 2023-10-03
Not enough char....


Ndoll1:
String for preferences and allergies is limited in character length. Is it possible to double the amount?

I wanted to add that I prefer dishes that are not vegan/vegetarian. A check would also work for me.

LenaWil:
Oh wait, you said double.  
I should have read closer.

Ndoll1:
1000 is also fine. But then check breaking styling bugs explicitly.

LenaWil:
I forgot to close this.

PR #267, opened 2023-10-03
Change the allergies length

LenaWil:
It’s literally an one character change, so that’s why I did this one and not the other. (yet) Not sure if this the correct amount, also I have no idea if this is enough/too much.

#265  @Ndoll1


Ndoll1 (commented):
Code looks fine. Check for length breaking bugs. So maybe a Lorem ipsum of 1000 characters?

mhvis (approved):
Looks good, thanks! I like the other ideas too (the meat/vega things).

LenaWil:
> Code looks fine. Check for length breaking bugs. So maybe a Lorem ipsum of 1000 characters?

I did do a quick and dirty test and it seemed to be fine.

PR #275, opened 2023-12-28
Change locale formatting in preparation for Django 5.0

mhvis:


mhvis:
This PR has as a side effect that the number formatting is changed from a dot as decimal separator to a comma.

So a balance value that looked like 4,431.50 before now looks like 4.431,50.

That's because the built-in Dutch locale in Django is setup that way.

mhvis:
Also I changed the date formatting like this

image

creditmanagement/templatetags/credit_tags.py
```diff
@@ -1,15 +1,14 @@
...
 register = template.Library()
 
 
 @register.filter
 def euro(value):
     """Format for euro values."""
-    amount = "{:.2f}".format(abs(value))
-    v = "{}€{}".format("-" if value < 0 else "", amount)
-    return mark_safe(v)
+    # Could surround this with `mark_safe` but is probably not necessary
```
DutcherNL, line 10: I see no reason why it's here either (except for the euro sign, but that should work without making it safe. After the update the comment is not neccessary though, so maybe just remove that line
```suggestion
```

DutcherNL (approved):

creditmanagement/templatetags/credit_tags.py
mhvis: Yes good point. I'll remove it.

PR #282, opened 2024-01-05
Comment e-mails, pinning and deletion

mhvis:
Also simplify the models in admin and change the dining list overview.

image

image


mhvis:
Dining list owners can send an e-mail to all diners.

Others can send an e-mail to the dining list owners. Also if they not joined the dining list, for instance to ask if they can still join when it's closed.

The sender e-mail address is added as Reply-To.

assets/templates/mail/comment_notification.txt
```diff
@@ -0,0 +1,10 @@
+{% load settings %}
+
+{{ comment.poster }} posted the following comment:
+
+{{ comment.message }}
+
+You can respond to {{ comment.poster }} by replying to this e-mail.
```
DutcherNL, line 7: Damn, nice function. How did you set that up?

general/mail_control.py
```diff
@@ -8,6 +8,10 @@
 
 from userdetails.models import User
 
+"""
+Deprecated: the method User.send_email() is simpler and can replace the functions in this module.
+"""
```
DutcherNL, line 13: Understandable, though I will miss the small html markup that was used on the emails

DutcherNL (approved):

general/mail_control.py
mhvis: Don't worry, I won't get rid of the nice HTML emails that we already have :smile:. I'm just not bothering with writing them myself. But if you want to add them I won't mind.

It's just that HTML emails are so difficult to write because they render differently in different e-mail clients. So you need to test them in all major clients. And short to the point e-mails are also easier to read.

assets/templates/mail/comment_notification.txt
mhvis: I'm not doing fancy stuff :blush:, just keeping it simple. The sender e-mail is added in the Reply-To header [here](https://github.com/VSEScala/Scala-Dining/pull/282/files#diff-2608dfa2b591f048a4302446578a26fd55e8062c86e25c6b1cdafd20eed71d98R587).
It's mentioned under the checkbox that the sender's e-mail will be included in the message, see the screenshot. That also kind of acts as a disincentive for spam, because the receiver gets to know the sender's e-mail.

Although I just realize that they could use a unverified e-mail address on the site. But as long as that's not being misused we can keep it like this.

mhvis:
I changed the comments order in this PR, to newest comments first. But now I realize that oldest first like how it was is probably better. So I'm going to revert it back to oldest first.

PR #283, opened 2024-01-13
Add kind of dish (meat/fish/vegetarian/vegan)

### LINKED ISSUE #266: Meat, vegetarian, vegan icon visible to indicate what kind of dish is prepared
Add an icon to notify eaters if list is Vegetarian, Vegan or contains meat (maybe also fish as an extra option). 

A lot of times list don't state if it contains meat or if it is vegetarian/vegan. That is unclear a lot of times. When cooking a person probably already knows if a list will be vegetarian/vegan or if it will contain meat before even deciding what to cook specifically.

Can be added here
image

Or in the information overview of the list in the list overview
image

mhvis:
I added a 'kind of dish' field. What do you think @Ndoll1? If you have ideas for additional kinds or other fields lmk (maybe lactose-free/gluten/..). Thanks for creating #266.

Overview:
image

Dining list page:
image

Choices:
image

Closes #266


Ndoll1:
Looks great. 

In the EU there is a list of allergens that need to be specified in restaurants and need to be bold on packages. 

image

If you want to add something else look into those. Maybe as allergensoverview for the cooks with English and Dutch names of what they need to check for.

Ndoll1 (approved):

mhvis:
I was thinking about something like that indeed. But it might be quite a lot of work for the cook to specify all nutrients. And there's a risk of errors. So it's probably better to have people with (difficult) allergies write a comment.

Ndoll1:
That is true. I meant it in a way that anyone can preselect these allergens as I have that allergy. And then other as comment. (Personal information)

The cook can then get on overview of all allergens. Is easier to check in my opinions. Also since it is a preselect you can provide the info in multiple languages. Because the packages not always contains English info, but Dutch they do.

mhvis:
Oh yeah that is smart. I was thinking the other way around, but it's indeed much better to just improve the allergens field in the personal account information to have a selection of choices.

Ndoll1:
Yep. But keep in mind that other is really important then as well. These 14 do not contain all allergens. 

I also would advise to separate allergies and wishes then.

PR #287, opened 2024-01-15
Fix dining info change bug and remove total dining cost field

mhvis:
I removed the total dining cost field because it seems to be confusing and it is not saved on the dining list anyway.


mhvis:
I'm merging this now because there was another error report today caused by this bug.

PR #289, opened 2024-01-18
Two additional options for kind of dish

### LINKED ISSUE #288: Change "Kind of dish" option to multi-options instead of a single option
Today I noticed this new feature (very nice addition) but I also noticed that you can only pick one of the options while some dishes might feature options. For example when making an AVG meal one might make vega meat for the vega's and real meat for the normal people.

So my suggestion is to impliment a checkbox there (instead of a single option) (or to change the name into "Vegetarian suitable").

mhvis:
I like the two extra options. Now I have it like this: 9baa1aebe11cda99df5a0a1b0eb96895a7e8dd08

Closes #288

### DIFF
--- a/dining/models.py
+++ b/dining/models.py
@@ -58,10 +58,12 @@ class DiningList(models.Model):
         blank=True,
         choices=[
             ("", "Not specified"),
-            ("meat", "🍗 Contains meat"),
-            ("fish", "🐟 Contains fish"),
+            ("meat", "With meat"),
+            ("fish", "🐟 With fish"),
             ("vegetarian", "🥕 Vegetarian"),
             ("vegan", "🌿 Vegan"),
+            ("meat_or_vegetarian", "Meat or vegetarian"),
+            ("meat_or_vegan", "Meat or vegan"),
         ],
     )
 
--- a/dining/models.py
+++ b/dining/models.py
@@ -58,7 +58,7 @@ class DiningList(models.Model):
         blank=True,
         choices=[
             ("", "Not specified"),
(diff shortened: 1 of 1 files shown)


dining/models.py
Ndoll1, line 1: Please add 🍗 for consistency

Ndoll1 (changes requested):
Please add 🍗 for consistency at "With meat"

Maybe also at  
🍗Meat or 🥕 Vegetarian
🍗Meat or 🌿 Vegan

Would still prefer to instead of "🍗Meat or 🥕 Vegetarian" and "🍗Meat or 🌿 Vegan". It would give an option as optional checkbox
🥕 Vegetarian welcome/friendly
🌿 Vegan welcome/friendly

Ndoll1 (approved):

mhvis:
:+1: Added the meat icon. I don't want too much of the icons because they're not very accurate actually :laughing:. Since a carrot is both vegetarian and vegan, and meat isn't necessarily a poultry leg.

mhvis:
> Would still prefer to instead of "🍗Meat or 🥕 Vegetarian" and "🍗Meat or 🌿 Vegan". It would give an option as optional checkbox 🥕 Vegetarian welcome/friendly 🌿 Vegan welcome/friendly

This suggests to me more like the meal is with meat by default but has a vegetarian option. Instead of you can choose between meat and vegetarian. That's also fine by me if that's a better way of doing it. But I want someone else to agree that this is better before I change it, otherwise lets leave it like this.

Issue #295, opened 2024-12-19
Create a filter in the admin panel for private dining lists


thomcrielaard:
The treasurer needs to send an invoice to private dining lists such that they are invoiced as a full dining list. Currently there is no way (I could find) to get a quick overview of private dining lists.

mhvis:
Thanks! Het staat online

PR #296, opened 2024-12-19
Added filter for association only in admin panel

thomcrielaard:
#295 
I also renamed the title of the field to be "Association only" instead of "Limit signups to association only" to keep a more concise overview. Since there is a description field added anyways, this should be enough information for the user.


mhvis:
Thanks!!

